PR #9, opened 2026-06-15
Remove unused dependencies and imports

glbrntt:
Motivation:

`NIOHTTP1` is a target dependency but is never used. `NIOPosix` is imported
but isn't necessary (it seems to be imported for it's transitive import of 
`CNIOLinux` to get various errno values). Integration tests depend on `NIOEmbedded`
but don't declare it as a dependency.

Modifications:

- Remove NIOHTTP1 target dependency
- Remove NIOPosix imports and add various lib C imports
- Add NIOEmbedded target dependency

Result:

More accurate dependency list

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -119,6 +119,7 @@ let package = Package(
                 .product(name: "NIOConcurrencyHelpers", package: "swift-nio"),
                 .product(name: "NIOCore", package: "swift-nio"),
                 .product(name: "NIOPosix", package: "swift-nio"),
+                .product(name: "NIOEmbedded", package: "swift-nio"),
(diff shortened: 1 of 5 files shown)


josephnoir:
Looks like it's missing `NIOPosix` now.

glbrntt:
> Looks like it's missing `NIOPosix` now.

`NIOPosix` shouldn't have been necessary. Looks like it was used for the transitive `CNIOLinux` import for errno values. I've replaced that with libc imports.

Also the integration tests import `NIOEmbedded` but didn't declare it in the manifest so I added it there as well.

josephnoir (approved):
LGTM!

PR #12, opened 2026-06-16
Replace stream-state `Bool` flags with state machines

rnro:
### Motivation

* `QUICStreamStateMachine` carried raw `Bool` flags `initialized` and `detached`. Flat `Bool`s lose distinctions like `.initializing` vs `.initialized`.
* `initialized` hid a race: concurrent reads on the same event-loop tick could both pass the gate before the initializer's `Future` flipped the flag.
* `lowerProtocol` was a sibling field reassigned to a stub on detach, so callers silently no-op'd on a dead reference.

### Modifications

* Replace `initialized` with a peer `QUICStreamPipelineStateMachine` tracking the channel pipeline initializer's lifecycle.
* Replace `detached` plus the `lowerProtocol` field with a `SwiftNetworkStreamHandle` wrapper that owns the linkage behind a nested state machine; callsites become single method calls instead of switching on attachment state inline.
* Lift the per-case inbound-init logic out of `QUICConnectionChannelHandler.channelRead` into `initialize(...)` overloads on `QUICChannelStreamHandler`, collapsing the nested `Future`-handling into one-liner dispatch.
* Add SM tests covering the meaningful transitions.

### Result

* Complexity and decision-making are centralized, not spread amongst different methods/call-sites.
* `QUICStreamStateMachine` is purely RFC 9000 again; pipeline and SwiftNetwork concerns sit beside it on the handler.
* Concurrent reads can no longer double-dispatch the initializer.
* Linkage invocations after detach are compiler-enforced.
* `.initializing` is observably distinct from `.initialized`, as the half-closure `receiveResetStream` deferral path needs.

### DIFF
--- a/Sources/NIOQUIC/QUICConnectionChannelHandler.swift
+++ b/Sources/NIOQUIC/QUICConnectionChannelHandler.swift
@@ -237,12 +237,7 @@ extension QUICConnectionChannelHandler: ChannelInboundHandler, ChannelOutboundHa
             return
         }
 
-        guard stream.isStreamChannelActive else {
-            stream.streamRead()
-            return
-        }
-
-        switch stream.pipelineStateMachine.startInitializer() {
+        switch stream.pipelineStateMachine.startInitializer(channelActive: stream.isStreamChannelActive) {
         case .runInitializer:
             self.logger.trace(
                 "QUICConnectionChannelHandler read with newly created id: \(readableStreamMessage.streamID)"
@@ -256,10 +251,7 @@ extension QUICConnectionChannelHandler: ChannelInboundHandler, ChannelOutboundHa
                 stream.initialize(context)
             }
 
-        case .skipInitializerInProgress:
-            stream.streamRead()
-
-        case .skipInitializerComplete:
+        case .ignore:
             stream.streamRead()
         }
     }
--- a/Sources/NIOQUIC/QUICConnectionChannelHandler.swift
+++ b/Sources/NIOQUIC/QUICConnectionChannelHandler.swift
@@ -236,42 +236,30 @@ extension QUICConnectionChannelHandler: ChannelInboundHandler, ChannelOutboundHa
             self.logger.info("Could not find stream: \(readableStreamMessage.streamID)")
             return
         }
-        // Make sure the stream channel is active
-        if !stream.streamStateMachine.initialized && stream.isStreamChannelActive {
+
+        guard stream.isStreamChannelActive else {
+            stream.streamRead()
+            return
+        }
+
+        switch stream.pipelineStateMachine.startInitializer() {
+        case .runInitializer:
             self.logger.trace(
                 "QUICConnectionChannelHandler read with newly created id: \(readableStreamMessage.streamID)"
             )
             switch self.inboundStreamInitializer {
             case .multiplexer(let continuation):
-                let f = continuation.initialize(channel: stream, streamID: readableStreamMessage.streamID)
-                f.whenComplete { result in
-                    switch result {
-                    case .success(let output):
-                        stream.streamStateMachine.initialized = true
-                        continuation.yield(output: output, channel: stream)
-                        stream.streamRead()
-                    case .failure:
-                        stream.close(promise: nil)
-                    }
-                }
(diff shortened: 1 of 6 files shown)


Sources/NIOQUIC/SwiftNetwork/QUICStreamStateMachine.swift
```diff
@@ -2030,3 +2028,383 @@ extension QUICStreamReceiveStateMachine.State {
...
+            self = .initialized
+            return .surfaceInitializedStream
+        case .initialized:
+            self = .initialized
+            return .ignoreAlreadyComplete
+        }
+    }
+}
+
+/// Handle a QUIC stream uses to talk to the SwiftNetwork transport stack.
```
josephnoir:
```suggestion
/// The handle a QUIC stream uses to talk to the SwiftNetwork transport stack.
```
Nit: Read "handle" as a verb first.

Sources/NIOQUIC/SwiftNetwork/QUICStreamStateMachine.swift
```diff
@@ -2030,3 +2028,383 @@ extension QUICStreamReceiveStateMachine.State {
...
+        }
+
+        enum InvokeDisconnectAction: ~Copyable {
+            case performDisconnect
+            case skipDetached
+        }
+        func invokeDisconnect() -> InvokeDisconnectAction {
+            switch self.state {
+            case .attached: return .performDisconnect
+            case .detached: return .skipDetached
```
josephnoir: Can we leave the "Detached" out of `skipDetached`? It does not help to clarify the action. `skipDisconnect` would be more accurate (or just `skip`). Same for the other actions in this state machine.

Sources/NIOQUIC/QUICConnectionChannelHandler.swift
```diff
@@ -236,42 +236,30 @@ extension QUICConnectionChannelHandler: ChannelInboundHandler, ChannelOutboundHa
...
-                stream.streamRead()
+                stream.initialize(context)
             }
-        } else {
+
+        case .skipInitializerInProgress:
+            stream.streamRead()
+
+        case .skipInitializerComplete:
             stream.streamRead()
```
josephnoir: Since these do the same thing, could the action returned here be one of `runInitializer` or `skipInitializer`?

Sources/NIOQUIC/QUICConnectionChannelHandler.swift
```diff
@@ -236,42 +236,30 @@ extension QUICConnectionChannelHandler: ChannelInboundHandler, ChannelOutboundHa
             self.logger.info("Could not find stream: \(readableStreamMessage.streamID)")
             return
         }
-        // Make sure the stream channel is active
-        if !stream.streamStateMachine.initialized && stream.isStreamChannelActive {
+
+        guard stream.isStreamChannelActive else {
+            stream.streamRead()
+            return
+        }
```
josephnoir: I'm stumbling over this check. Even if this is not the case, `stream.streamRead()` will still be called below without checking this condition again. That makes me wonder if the `isStreamChannelActive` check is actually cheaper than `stream.pipelineStateMachine.startInitializer()` and if we can just drop this.

josephnoir (commented):
Thanks Rick! I have a few small things I want to ask about.

Sources/NIOQUIC/QUICConnectionChannelHandler.swift
rnro: Yeah I wondered about this myself, I've moved this into the state machine's decision.

Sources/NIOQUIC/SwiftNetwork/QUICStreamStateMachine.swift
rnro: I've renamed this `.ignore` which is more common I think.

Sources/NIOQUIC/QUICConnectionChannelHandler.swift
rnro: Similar to the last one, I've renamed this `.ignore` with an associated reason enum.

josephnoir (approved):
LGTM!

PR #14, opened 2026-06-16
Remove incrementalMaxStreamsUpdate argument

aryan-25:
### Motivation

`QUICConfiguration`'s initializers accept an `incrementalMaxStreamsUpdate` argument, but this is not used anywhere: updates are handled by SwiftQUIC directly. We should therefore remove the `incrementalMaxStreamsUpdate` argument from `QUICConfiguration`.

### Modifications

Removed `incrementalMaxStreamsUpdate` from `QUICConfiguration`.

### Result

It becomes easier to initialize a `QUICConfiguration` instance.


josephnoir (approved):
Thanks!

glbrntt (approved):

PR #17, opened 2026-06-17
Adopt fix for static builds

josephnoir:
### Motivation

Linux static SDK builds are failing in CI due to a nettling dependency. A new branch in swift-network-evolution removes it to make the builds easier.

### Modifications

Switch swift-network-evolution reference.

### Result

Enable Linux static SDK builds

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -52,7 +52,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-nio-quic-helpers.git", branch: "main"),
         .package(
             url: "https://github.com/apple/swift-network-evolution",
-            revision: "19a73fc78ad5845a79ec67ebbb6cc3b0cf2e8f9b",
+            revision: "ca1af826afc6408a2a68eb795db978c16e5ced89",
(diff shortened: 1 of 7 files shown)


Package.swift
```diff
@@ -52,7 +52,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-nio-quic-helpers.git", branch: "main"),
         .package(
             url: "https://github.com/apple/swift-network-evolution",
-            revision: "ba18f3f82016d098f83de061b67f46c48f8639a1",
+            revision: "19a73fc78ad5845a79ec67ebbb6cc3b0cf2e8f9b",
```
glbrntt: [This just landed](https://github.com/apple/swift-network-evolution/commit/ca1af826afc6408a2a68eb795db978c16e5ced89) which should fix the flaky tests, might as well use that version here

```suggestion
            revision: "ca1af826afc6408a2a68eb795db978c16e5ced89",
```

glbrntt (approved):

PR #19, opened 2026-06-18
Pick up latest swift-network-evolution changes

rnro:
Pick up recent availability and import changes which should help (but not necessarily make work) Static SDK and watchOS.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -54,7 +54,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-nio-quic-helpers.git", branch: "main"),
         .package(
             url: "https://github.com/apple/swift-network-evolution",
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -54,7 +54,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-nio-quic-helpers.git", branch: "main"),
         .package(
             url: "https://github.com/apple/swift-network-evolution",
-            revision: "1b7579b056416cdc53d7a66dde4c498bf6983afe",
+            revision: "da2df73192a294646afbbfade4ade8881a28e18f",
```
rnro:
```suggestion
            revision: "ba18f3f82016d098f83de061b67f46c48f8639a1",
```

glbrntt (approved):

PR #22, opened 2026-06-24
Add AliasDictionary

glbrntt:
Motivation:

We plan to remove the child-channel-mux for connections in favour of a custom implementation. This will land as a number of PRs rather than in one big chunk.

This first part of this is a dictionary which supports key aliasing. This will in the fullness of time be used for storing connections keyed by their connection ID(s).

Modifications:

- Adds an `AliasDictionary`, where multiple keys resolve to the same value. Each value has a canonical key and zero or more aliases. Removing the canonical key promotes an alias if one exists, or removes the value if no aliases exist.

Result:

First piece in place for replacing connection mux


Sources/NIOQUIC/AliasDictionary.swift
```diff
@@ -0,0 +1,268 @@
...
+        }
+    }
+
+    /// Adds `alias` as another way to look up the value stored under `key`.
+    ///
+    /// `key` may be either the canonical key or an existing alias; in either case the new alias
+    /// is registered against the canonical key.
+    ///
+    /// - Precondition: `alias` must not already exist as a canonical key or alias in this
+    ///   dictionary.
```
josephnoir: Doesn't the second guard catch this?

josephnoir (approved):
Looks good! Just one question about the docs.

Sources/NIOQUIC/AliasDictionary.swift
glbrntt: Yes! I updated the code but not the docs, good catch.

glbrntt:
```suggestion
```

PR #23, opened 2026-06-24
Support stream-level backpressure

aryan-25:
### Motivation

Incoming stream data delivered to `QUICChannelStreamHandler`'s `handleInboundDataAvailableEvent` is immediately read out of SwiftQUIC's buffer and fired downstream. Since the internal buffer is being emptied every time we receive stream data from the network, the flow-control mechanisms aren't provided accurate signals, and therefore backpressure cannot effectively be exerted.

We should only drain the stream buffer when the downstream has requested a read in order to exert backpressure.

### Modifications

- Updated `QUICChannelStreamHandler`:
  - Added a `pendingRead` boolean flag (set to `false` by default).
  - Added a new private method named `deliverInboundDataDownstreamIfDemand`, which first checks whether `pendingRead == true`, and only then reads data out of SwiftQUIC's buffer.
  - Updated `handleInboundDataAvailableEvent` to call into `deliverInboundDataDownstreamIfDemand`.
  - Added support for the `autoRead` channel option.
    - This includes firing a `read` event on the pipeline after `channelReadComplete` if `autoRead == true`.
- Added associated tests.

### Result

The stream buffer is now emptied only when the downstream consumer requests a read. This will make SwiftQUIC more aware of whether the downstream is consuming quickly enough, and therefore provide more accurate input to its internal flow-control mechanisms.


Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -528,28 +544,62 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
...
+    private func _fireChannelReadComplete() {
+        self.log("ChildChannel fire channel read complete")
+        self.pipeline.fireChannelReadComplete()
+
+        self._tryToAutoRead()
+    }
+
+    /// Issues a `read()` on the pipeline if `autoRead` is enabled and there is no outstanding read request.
+    private func _tryToAutoRead() {
+        if self.autoRead, self.pendingRead == false {
```
glbrntt:
```suggestion
        if self.autoRead, !self.pendingRead {
```

Sources/NIOQUIC/QUICConnectionChannelHandler.swift
```diff
@@ -242,13 +242,33 @@ extension QUICConnectionChannelHandler: ChannelInboundHandler, ChannelOutboundHa
...
-            switch self.inboundStreamInitializer {
-            case .multiplexer(let continuation):
-                stream.initialize(multiplexerContinuation: continuation, streamID: readableStreamMessage.streamID)
-            case .closure(let initializer):
-                stream.initialize(context, initializer)
-            case .none:
-                stream.initialize(context)
+
+            let loopBoundSelf = NIOLoopBound(self, eventLoop: self.eventLoop)
+            let loopBoundContext = NIOLoopBound(context, eventLoop: self.eventLoop)
```
glbrntt: Don't use loop bound here, it's not necessary. Prefer using the `assumeIsolated()` variants for futures so that you can attach callbacks which don't require `@Sendable`.

Sources/NIOQUIC/QUICConnectionChannelHandler.swift
```diff
@@ -242,13 +242,33 @@ extension QUICConnectionChannelHandler: ChannelInboundHandler, ChannelOutboundHa
...
+
+            // Inherit the parent channel's `autoRead` option onto the stream channel.
+            let setStreamAutoRead = context.channel.getOption(.autoRead).flatMapThrowing { autoRead in
```
glbrntt: You should implement the sync options path as well. Most channels will implement it and it avoids a bunch of futures and callbacks. You'll still need the async path as a fallback.

Sources/NIOQUIC/QUICConnectionChannelHandler.swift
```diff
@@ -242,13 +242,33 @@ extension QUICConnectionChannelHandler: ChannelInboundHandler, ChannelOutboundHa
...
+            setStreamAutoRead.whenSuccess {
+                switch loopBoundSelf.value.inboundStreamInitializer {
+                case .multiplexer(let continuation):
+                    stream.initialize(multiplexerContinuation: continuation, streamID: readableStreamMessage.streamID)
+                case .closure(let initializer):
+                    stream.initialize(loopBoundContext.value, initializer)
+                case .none:
+                    stream.initialize(loopBoundContext.value)
+                }
             }
```
glbrntt: combine these in `whenComplete` otherwise you're paying the cost of two callbacks

Sources/NIOQUIC/QUICConnectionChannelHandler.swift
```diff
@@ -447,14 +467,38 @@ extension QUICConnectionChannelHandler {
...
+
+        setStreamAutoRead.whenSuccess {
+            loopBoundStreamInitializer.value(streamHandler, streamID)
+                .whenComplete { result in
+                    switch result {
+                    case .success:
+                        channelActivationPromise.succeed(streamHandler)
+                        // Run an empty `streamRead()` to fire `channelReadComplete` which will in turn fire a `read`
+                        // event on the pipeline.
+                        streamHandler.streamRead()
```
glbrntt: `tryToAutoRead` seems like the more appropriate function to call.

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -572,7 +622,7 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
             break
         }
         // Make sure read complete is called after everything has run.
-        self.pipeline.fireChannelReadComplete()
+        self._fireChannelReadComplete()
```
glbrntt: This isn't your bug, but we shouldn't fire read complete if we didn't fire read.

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -1272,3 +1334,29 @@ extension QUICChannelStreamHandler: Channel, ChannelCore {
...
+    internal var _testOnly_pendingRead: Bool {
+        self.eventLoop.preconditionInEventLoop()
+        return self.pendingRead
+    }
+
+    /// Obtain the `autoRead` property.
+    internal var _testOnly_autoRead: Bool {
+        self.eventLoop.preconditionInEventLoop()
+        return self.autoRead
+    }
```
glbrntt: do we really need this when we have sync options?

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -1272,3 +1334,29 @@ extension QUICChannelStreamHandler: Channel, ChannelCore {
...
+        self.eventLoop.preconditionInEventLoop()
+        var buffer = buffer
+        self.bufferedReadData.writeBuffer(&buffer)
+    }
+
+    /// Obtain the `pendingRead` property.
+    internal var _testOnly_pendingRead: Bool {
+        self.eventLoop.preconditionInEventLoop()
+        return self.pendingRead
+    }
```
glbrntt: Do we need this? Can't we observe the behavior when calling `handleInboundDataAvailableEvent` instead?

Tests/NIOQUICTests/QUICChannelStreamHandlerTests.swift
```diff
@@ -0,0 +1,475 @@
...
+                    return accumulated + buffer.readableBytes
+
+                default:
+                    return accumulated
+                }
+            }
+        }
+
+        var channelReadCount: Int {
+            self.events.filter {
```
glbrntt: nit: use `count(where:)`

Tests/NIOQUICTests/QUICChannelStreamHandlerTests.swift
```diff
@@ -0,0 +1,475 @@
...
+
+        func read(context: ChannelHandlerContext) {
+            self.events.append(.read)
+
+            // Don't call context.read() now; create a promise that will call `context.read()` and store that promise.
+            // That promise will be fulfilled when `releasePendingReadRequest` is called. At that point, the read will
+            // be propagated further down the pipeline.
+            let readPromise = context.eventLoop.makePromise(of: Void.self)
+            let loopBoundContext = context.loopBound
+            readPromise.futureResult.whenComplete { _ in
```
glbrntt: Don't use loop bound, use `assumeIsolated` here

Tests/NIOQUICTests/QUICChannelStreamHandlerTests.swift
```diff
@@ -0,0 +1,475 @@
...
+                loopBoundContext.value.read()
+            }
+            self.pendingReadRequests.append(readPromise)
+        }
+
+        /// Release a read request we received previously and propagate it down the pipeline.
+        func releasePendingReadRequest(to pipeline: ChannelPipeline) {
+            guard let promise = self.pendingReadRequests.popLast() else { return }
+            promise.succeed()
+        }
```
glbrntt, line 445: This handler isn't really a recording handler, it's a recording handler and a hold-up-the-reads-until-I-release-them handler. Either it's badly named or it should be two separate handlers.

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -1272,3 +1334,29 @@ extension QUICChannelStreamHandler: Channel, ChannelCore {
...
+//
+// These methods are intended to only be used in tests. They exist to let unit tests drive the inbound delivery path
+// without a live SwiftNetwork peer.
+@available(anyAppleOS 26, *)
+extension QUICChannelStreamHandler {
+    /// Append bytes directly into the NIO read buffer.
+    internal func _testOnly_appendToBufferedReadData(_ buffer: ByteBuffer) {
+        self.eventLoop.preconditionInEventLoop()
+        var buffer = buffer
+        self.bufferedReadData.writeBuffer(&buffer)
```
glbrntt: just use `writeImmutableBuffer`

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -1272,3 +1334,29 @@ extension QUICChannelStreamHandler: Channel, ChannelCore {
         fatalError("not implemented \(#function)")
     }
 }
+
+// MARK: - Test accessors
+//
+// These methods are intended to only be used in tests. They exist to let unit tests drive the inbound delivery path
```
glbrntt: I generally don't like adding hooks for tests. Observing behavior is typically much better than checking internal state (a change in state doesn't mean we get the desired behavior).

For appending the buffered data, fine, you're avoid a real network peer. The other two shouldn't be necessary though.

glbrntt (changes requested):
Thanks Aryan -- this mostly looks good, I left some notes inline.

agnosticdev (commented):
Can we run a before and after performance test with this also? No argument that we need this I just want to double check if there is a performance impact. I’d rather address it now than afterwards.

Sources/NIOQUIC/QUICConnectionChannelHandler.swift
aryan-25: Addressed in [8442365](https://github.com/apple/swift-nio-quic/pull/23/commits/84423656ae70e2dd06141dda364107f0996df1c3).

Tests/NIOQUICTests/QUICChannelStreamHandlerTests.swift
aryan-25: Agreed. I've split this into two handlers in [89eddd9](https://github.com/apple/swift-nio-quic/pull/23/commits/89eddd9cbd5d6f835575d09881daa76d2576ecee).

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
aryan-25: Thanks for this, I've removed this property now and replaced all its references with `syncOptions.getOption(.autoRead)`.

Sources/NIOQUIC/QUICConnectionChannelHandler.swift
aryan-25: Addressed in [ab84a1c](https://github.com/apple/swift-nio-quic/pull/23/commits/ab84a1c38f39664143ea234bbcce53b36cc97292).

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
aryan-25: Addressed in [e1bf426](https://github.com/apple/swift-nio-quic/pull/23/commits/e1bf4267ae6f3399cb7c86207040fa401933edc6).

Sources/NIOQUIC/QUICConnectionChannelHandler.swift
```diff
@@ -447,14 +451,86 @@ extension QUICConnectionChannelHandler {
...
+    private func inheritAutoRead(
+        syncOptions: any NIOSynchronousChannelOptions,
+        streamChannel: QUICChannelStreamHandler
+    ) -> Result<Void, any Error> {
+        Result(catching: {
+            let autoReadValue = try syncOptions.getOption(.autoRead)
+
+            // Force unwrap is safe here. `QUICChannelStreamHandler` always provides `syncOptions`.
+            try streamChannel.syncOptions!.setOption(.autoRead, value: autoReadValue)
+        })
```
glbrntt: Trailing closure syntax should work here.

```suggestion
        Result {
            let autoReadValue = try syncOptions.getOption(.autoRead)

            // Force unwrap is safe here. `QUICChannelStreamHandler` always provides `syncOptions`.
            try streamChannel.syncOptions!.setOption(.autoRead, value: autoReadValue)
        }
```

Sources/NIOQUIC/QUICConnectionChannelHandler.swift
```diff
@@ -447,14 +451,86 @@ extension QUICConnectionChannelHandler {
...
+
+    /// Synchronously reads the parent channel's `autoRead` option via `syncOptions` and applies it to `streamChannel`.
+    private func inheritAutoRead(
+        syncOptions: any NIOSynchronousChannelOptions,
+        streamChannel: QUICChannelStreamHandler
+    ) -> Result<Void, any Error> {
+        Result(catching: {
+            let autoReadValue = try syncOptions.getOption(.autoRead)
+
+            // Force unwrap is safe here. `QUICChannelStreamHandler` always provides `syncOptions`.
```
glbrntt, line 518: It's true that `QUICChannelStreamHandler` provides `SyncOptions`, however this is a `ChannelHandler`, so it's entirely possible for it to get added to the pipeline for a different channel type (e.g. `EmbeddedChannel` in tests). 

In practice it's fine because the handler isn't `public` and we control what channels it can be added to, so this is just a cautionary note.

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
glbrntt: This one didn't get resolved or responded to.

Sources/NIOQUIC/QUICConnectionChannelHandler.swift
glbrntt: Ah, sorry, this being an extension in `QUICConnectionChannelHandler` threw me.

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
aryan-25: Addressed in [5904ad0](https://github.com/apple/swift-nio-quic/pull/23/commits/5904ad0254a78e3a9eb879d4eb17e8691e1e012c).

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
aryan-25: Thanks for this, I've removed the other two properties now.

glbrntt (approved):
Looks great, thanks @aryan-25!

PR #31, opened 2026-06-26
Add CI job for HTTP/3 integration tests

rnro:
### Motivation

swift-nio-http3 is tightly coupled with this repository and it is important that we don't regress it during active 0.x development.

### Modifications

Clone and run swift-nio-http3 integration tests on PRs and main.

### Result

Regressions are spotted quicker


rnro:
The new jobs seem to be correctly finding the dependency mismatch with HTTP/3 I have seen locally.

glbrntt (approved):

PR #32, opened 2026-06-26
Add QUICConnectionChannel core parts

glbrntt:
Motivation:

We're adding a custom connection channel to make connection multiplexing cheaper. This lands some (but by no means all) of its internals.

Modifications:

Add QUICConnectionChannel with:
- Its stored state and init
- A lifecycle state machine
- Channel and ChannelCore conformance
- close, drain, and channel-inactive plumbing

Result:

Not much, yet, as nothing uses this.

### DIFF
--- a/Sources/NIOQUIC/QUICConnectionChannel+Lifecycle.swift
+++ b/Sources/NIOQUIC/QUICConnectionChannel+Lifecycle.swift
@@ -1,0 +1,93 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2026 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+@available(anyAppleOS 26, *)
+extension QUICConnectionChannel {
+    /// The lifecycle state of a ``QUICConnectionChannel``.
+    enum Lifecycle: Hashable {
+        case idle
+        case initializing
+        case initialized
+        case activated
+        case closing
+        case closed
+
+        mutating func initialize() -> Bool {
+            let canInitialize: Bool
+            switch self {
+            case .idle:
+                self = .initializing
+                canInitialize = true
+            case .initializing, .initialized, .activated, .closing, .closed:
+                canInitialize = false
+            }
+            return canInitialize
+        }
+
+        mutating func initialized() {
+            switch self {
+            case .initializing:
+                self = .initialized
+            case .idle, .initialized, .activated:
+                fatalError("Internal inconsistency")
+            case .closing, .closed:
+                // The channel was closed while the user's initializer was still in flight.
+                // Don't transition; we're already on the close path.
+                break
+            }
+        }
+
+        mutating func activated() {
+            switch self {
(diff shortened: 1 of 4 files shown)


Sources/NIOQUIC/QUICConnectionChannel.swift
```diff
@@ -0,0 +1,416 @@
...
+    // @unchecked because of the IUO ChannelPipeline, which is never mutated after `init`.
+    // The `ChannelPipeline` breaks the retain cycle between it and this channel.
+
+    /// The pipeline associated with this channel.
+    private var _pipeline: ChannelPipeline!
+
+    /// Completed when the channel is closed. Provides `closeFuture` for the `Channel` API.
+    private let closePromise: EventLoopPromise<Void>
+
+    /// The underlying SwiftNetowork-backed QUIC connection.
```
josephnoir:
```suggestion
    /// The underlying SwiftNetwork-backed QUIC connection.
```

Sources/NIOQUIC/QUICConnectionChannel+Lifecycle.swift
```diff
@@ -0,0 +1,93 @@
...
+        mutating func closed() {
+            switch self {
+            case .idle, .initializing, .initialized, .activated, .closing:
+                self = .closed
+            case .closed:
+                break
+            }
+        }
+
+        func outboundDataProcessed() -> Bool {
```
josephnoir, line 82: This is the only one that doesn't do a state change. Maybe this could be reflected in the name and simply be called `isInitializing`?

Sources/NIOQUIC/QUICConnectionChannel.swift
```diff
@@ -0,0 +1,416 @@
...
+        connection: SwiftNetworkQUICConnection
+    ) {
+        self.udpChannel = udpChannel
+        self.udpView = udpView
+        self.connection = connection
+
+        self.eventLoop = udpChannel.eventLoop
+        self.allocator = udpChannel.allocator
+        self.closePromise = udpChannel.eventLoop.makePromise()
+        self._isWritable = Atomic(true)
```
josephnoir, line 71: Is this always `true`?

Sources/NIOQUIC/QUICConnectionChannel.swift
```diff
@@ -0,0 +1,416 @@
...
+        }
+
+        // The view discards unnecessary flushes; no need to track them per-write in the loop.
+        self.udpView.flush()
+
+        let isInitializing = self.state.lifecycle.outboundDataProcessed()
+        switch self.connection.outboundDataProcessed(isChannelInitializing: isInitializing) {
+        case .completeActivation:
+            self.state.lifecycle.activated()
+            self._isActive.store(true, ordering: .releasing)
```
josephnoir, line 354: Since this should only change on state changes, could this be part of the lifecycle?

josephnoir (commented):
Do you have more tests upcoming for the `QUICConnectionChannel` itself? There are tests here for the Lifecycle, but the channel implements logic for closing, sending, and events. It also drives its internal lifecycle, which might be worth a test.

glbrntt:
> Do you have more tests upcoming for the `QUICConnectionChannel` itself? There are tests here for the Lifecycle, but the channel implements logic for closing, sending, and events. It also drives its internal lifecycle, which might be worth a test.

No, but I think you've hit on a deeper problem there. This (and the existing impl.) are strongly coupled to `SwiftNetworkQUICConnection` which makes anything other than integration tests really difficult. I'm going to look into this... I suspect we can do some simplification.

PR #33, opened 2026-06-29
Fix `channelInactive` propagation hang and `SwiftNetworkStreamHandle`…

rnro:
### Motivation

  * `handleDisconnectedEvent` deferred its close cascade through `eventLoop.execute`, so `_closePromise.succeed()` landed on tick 2 - after `ChildChannelMultiplexer.removeHandlers` nilled the context on tick 1, silently dropping `fireChannelInactive` and hanging swift-nio-http3 async tests.
  * Running the close inline trades that hang for a Swift exclusivity trap: SwiftNetwork's `handleCallFromUpperProtocol` synchronously drains queued peer events in a `defer`. A disconnect arriving mid-`invoke*` re-enters `handleDisconnectedEvent`; an inline `invokeDetach` overlaps the outer read borrow on `self.swiftNetworkStreamHandle`.

  ### Modifications

  * `SwiftNetworkStreamHandle.invoke*` wrappers now return an action carrying the linkage by value. Callers run the SwiftNetwork side effect on the local copy, so no borrow is held during the call - synchronous re-entry into the handle is safe.
  * `handleDisconnectedEvent` runs cleanup synchronously: `invokeDetach`, then `_close` when `isActive`. `.skipAlreadyDetached` falls through to the `isActive` check so `_closePromise` still resolves on an already-detached stream.
  * `writeBuffersToStream`/`sendFIN` finalise their `FrameArray` as failed in the new violation arms — the wrapper no longer hands the array to the linkage there.
  * Regression tests in `QUICChannelStreamHandlerTests` cover the synchronous `_isActive` flip, the borrow-release contract, and `_closePromise` resolution on an already-detached stream.

  ### Result

  * Async H/3 integration tests pass; `testResetStream` (and similar write-then-close-with-peer-disconnect flows) no longer trap.
  * No `eventLoop.execute` on any path that touches `swiftNetworkStreamHandle` — the fix is structural at the wrapper boundary, not a hop-on-tick workaround.
  * The remaining `eventLoop.execute` wrapping the pipeline-state teardown is unchanged; `ChildChannelMultiplexer` re-entrancy is the deeper root cause, tracked separately.

### DIFF
--- a/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
+++ b/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
@@ -369,7 +369,12 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
         self.fromExternal {
             do throws(NetworkError) {
                 self.log("abortInbound")
-                try self.swiftNetworkStreamHandle.invokeAbortInbound(from: self.reference, error: error)
+                switch self.swiftNetworkStreamHandle.invokeAbortInbound() {
+                case .proceed(let linkage):
+                    try linkage.invokeAbortInbound(self.reference, error: error)
+                case .ignore:
+                    break
+                }
             } catch {
                 self.log("Failed to abort inbound: \(error)")
                 self.closeStream(mode: .disconnectOnly, error: error, promise: nil)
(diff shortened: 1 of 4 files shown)


glbrntt (approved):
Thanks Rick!

PR #34, opened 2026-06-30
Remove max log level trait

### LINKED ISSUE #30: Max log level trait is passed on to dependent packages
I’m not sure if this is a SwiftPM bug, but when this package is added as a dependency in other packages, the Swift Log trait `"MaxLogLevelDebug"`, enabled by this package, is included in the dependent package, and I'm unable to remove it, even by setting `traits: [.defaults]` or `traits: []`. This means that my package loses access to all lower log levels.

Again, I don’t know if this is a bug in SwiftPM, nor whether there’s a need to keep this trait enabled in this library, but if it could be removed, I think it would be much appreciated. Thanks in advance!

fpseverino:
### Motivation

Resolves #30 

### Modifications

Removes the max log level package trait from the Swift Log dependency

### Result

End-users can control the max log level

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -29,11 +29,6 @@ if isRunningInCI {
     qlogSetting = []
 }
 
-// controls logs emitted, lower levels are compiled out
-// `[.init(stringLiteral: "MaxLogLevelNone")]` gives best performance
-// `[]` gives all log output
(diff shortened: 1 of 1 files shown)


glbrntt (approved):
Thank you!

PR #36, opened 2026-07-01
Make SwiftTLS/Authenticator argument order consistent

aryan-25:
### Motivation

SwiftTLS/Authenticator has two initializers. One accepts the certificate chain and private key from filepaths, and the other from in-memory instances.

The file-path initializer takes the certificate chain first, whereas the in-memory initializer takes the private key first. The two orders should be consistent.

### Modifications

Reordered the arguments of the in-memory initializer to be consistent with the other initializer.

### Result

Better consistency in SwiftTLS/Authenticator initializers.


josephnoir (approved):

josephnoir:
API breakage is expected.

PR #37, opened 2026-07-01
Stream receive refactors

rnro:
### Motivation

* Preparing for future half-closure-correctness PRs
* `applicationRead` is used for its side effects in `completeRead` which is semantically unclear.
* `QUICStreamReceiveStateMachine.wasReset` may be easily conflated with `QUICStreamSendStateMachine.wasReset` at usage sites.

### Modifications

* Two pure refactoring commits:
  * Extract code for flushing read data and surfacing completions (this will soon be re-used)
  * Extract code into a method to workaround a compiler crash
* Introduce `consumeReceiveTerminator` in `completeRead` which is explicit about its intent meaning we don't need to handle inappropriate returned actions.
* `QUICStreamReceiveStateMachine.wasReset` renamed to `hasReceivedReset`, parallel to `hasReceivedFin`.

### Result

* Clearer code which will make for clearer half-closure PR reviews.

### DIFF
--- a/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
+++ b/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
@@ -266,29 +266,7 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
                 UInt64(applicationErrorCode)
             )!
             do {
-                switch try self.streamStateMachine.receiveResetStream(
-                    applicationErrorCode: errorCode,
-                    finalSize: 0
-                ) {
-                case .closeStream(_):
-                    self.closeStream(
-                        mode: .closeAndDisconnect,
-                        error: NIOQUICHelpers.QUICStreamResetError(code: errorCode),
-                        promise: nil
-                    )
-                case .doNotCloseStream(_):
-                    self.closeStream(
-                        mode: .closeAndDisconnect,
-                        error: NIOQUICHelpers.QUICStreamResetError(code: errorCode),
-                        promise: nil
-                    )
-
-                case .ignore(.alreadyFullyReceived):
-                    self.log("receiveResetStream: all data already received, ignoring")
-
-                case .ignore(.alreadyReset):
-                    self.log("receiveResetStream: stream already reset, ignoring")
-                }
+                try self.handleReceivedResetStream(applicationErrorCode: errorCode)
             } catch {
                 self.logger.warning("\(self.logPrefix) handleInboundAbortedEvent errored: \(error)")
                 switch error {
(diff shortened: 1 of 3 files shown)


Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -617,47 +625,51 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
...
+    }
+
+    func streamRead() {
+        self.eventLoop.preconditionInEventLoop()
+
+        // Emit channel read complete when:
+        // - any read was fired, or
+        // - EOF is emitted
+        let flushed = self.flushBufferedReadData()
+        let surfaced = self.surfaceReadCompletion()
```
glbrntt: `surfaced` isn't an obvious name here, `eof` maybe?

josephnoir (approved):

Sources/NIOQUIC/SwiftNetwork/QUICStreamStateMachine.swift
```diff
@@ -1103,6 +1106,41 @@ extension QUICStreamStateMachine.State {
         }
     }
 
+    /// Consuming a received terminator. Non-terminal or non-connected
+    /// states return `.nothingToReport`.
+    mutating func consumeReceiveTerminator()
```
glbrntt: This naming doesn't sit well; it feels jargon-y to me but I don't have a better alternative. However – this is only called in `completeRead`, can we inline the code there and remove some indirection?

glbrntt (changes requested):

Sources/NIOQUIC/SwiftNetwork/QUICStreamStateMachine.swift
rnro: So the indirection is there because I intend to call this from another site in an upcoming PR and I wanted the reviews to be easier since this is a straight refactor. Naming is tricky because I want to cover `FIN`('eof') and `RESET` cases, this will consume either.

glbrntt: > So the indirection is there because I intend to call this from another site in an upcoming PR and I wanted the reviews to be easier since this is a straight refactor.

Makes sense.

How about being more explicit, `consumeFinOrReset`?

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
rnro: I've changed it to `endOfStream` because similar to the other spot, this could be a `FIN` or ` RESET`.

glbrntt (approved):
Thanks – that's clearer!

PR #40, opened 2026-07-03
Store protocol reference

glbrntt:
Motivation:

The ProtocolInstanceReference on the QUICChannelStreamHandler is currently a computed property which does work every time its accessed. Just store it instead, ensuring it's nil'd out when tearing down the stream handler to break the retain cycle (as it has a ref to the stream handler).

Modifications:

- Create and store a ProtocolInstanceReference in init
- Break the retain cycle in close

Result:

A small perf improvement, 1-1.5%

### DIFF
--- a/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
+++ b/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
@@ -75,8 +75,8 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
     private let role: Role
     private let keepAliveInterval: Duration?
     private let logger: Logger
+    internal var reference: ProtocolInstanceReference = ProtocolInstanceReference()
 
-    internal var reference: ProtocolInstanceReference { ProtocolInstanceReference(custom: self) }
     internal var eventManager = ProtocolEventManager()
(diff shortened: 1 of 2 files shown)


josephnoir (approved):

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -75,8 +75,8 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
     private let role: Role
     private let keepAliveInterval: Duration?
     private let logger: Logger
+    internal var reference: ProtocolInstanceReference = ProtocolInstanceReference()
 
-    internal var reference: ProtocolInstanceReference { ProtocolInstanceReference(custom: self) }
```
agnosticdev, line 79: Yeah, this is a good idea.  Internally this is something we want to address over in SwiftNetwork too.  Thanks!

agnosticdev (approved):

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
glbrntt: Yeah, just need to be really careful about breaking the retain cycle!

PR #42, opened 2026-07-06
Implement half-closure for RESET_STREAM

rnro:
### Motivation

* NIO QUIC tore down the entire stream channel on a peer RESET_STREAM, even with `isOutboundHalfClosureEnabled: true`.
* Clean FIN and abrupt reset both surfaced as `ChannelEvent.inputClosed`.
* `QUICStreamStateMachine.initialized` was never set on outbound streams.

### Modifications

* When the receive side resets but the write side is still live, fire `errorCaught(QUICStreamResetError(code:))` on the inbound pipeline and leave the channel and outbound writer alone.
* Move decision-making into `QUICStreamStateMachine`.
* Mark outbound streams as initialized on outbound path.
* Add `anyErrorSeen` to `ErrorCatchingHandler` and update `testResetStream` / `testResetStreamRaceCondition` to await it instead of `closeFuture`, which no longer fires on RESET under half-closure.

### Result

* Applications can continue writing on a bidirectional stream after the peer resets its receive side.
* Peer resets surface as `errorCaught(QUICStreamResetError(code:))` and clean FINs as `ChannelEvent.inputClosed`.

**NOTE**: This PR leaves a timing window issue (to be closed in a follow up PR for cleaner reviews). The issue occurs when a `RESET_STREAM` arrives before initialization has completed. The receive SM captures it but only surfaces it to the application on the next read. Under `autoRead == false`, an application that never issues a read will never observe the reset.

### DIFF
--- a/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
+++ b/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
@@ -252,10 +252,13 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
     ///
     /// - Parameters:
     ///     - logMessage: The logMessage that is fetched by an autoclosure.  For performance reasons we could gate this behind a flag.
-    func log(_ logMessage: @autoclosure () -> String) {
+    func log(
+        _ logMessage: @autoclosure () -> String,
+        metadata: @autoclosure () -> Logger.Metadata? = nil
+    ) {
         #if DEBUG
         let message = logMessage()
-        self.logger.trace("\(self.logPrefix) \(message)")
+        self.logger.trace("\(self.logPrefix) \(message)", metadata: metadata())
         #endif
     }
 
@@ -265,7 +268,7 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
         error: SwiftNetwork.NetworkError?
     ) {
         if let error, let applicationErrorCode = error.quicApplicationError, applicationErrorCode >= 0 {
-            self.log("Received reset stream error: \(applicationErrorCode)")
+            self.log("Received reset stream error", metadata: ["applicationErrorCode": "\(applicationErrorCode)"])
             // applicationErrorCode comes from the wire as a QUIC variable-length
             // integer (< 2^62), so the force-unwrap is safe.
             let errorCode = NIOQUICHelpers.QUICApplicationErrorCode(
@@ -321,10 +324,10 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
     // Swift 6.3.
     @inline(never)
     private func surfaceReceivedReset(
-        applicationErrorCode code: NIOQUICHelpers.QUICApplicationErrorCode
+        applicationErrorCode code: QUICApplicationErrorCode
     ) {
         if self.pipelineStateMachine.isInitialized {
-            self.log("surfaceReset: applicationErrorCode=\(code)")
+            self.log("surfaceReset", metadata: ["applicationErrorCode": "\(code)"])
             // yields buffered data and then surfaces the reset
             self.streamRead()
         } else {
--- a/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
+++ b/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
@@ -307,7 +307,20 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
             )
 
         case .surfaceReset(let code):
+            #if compiler(<6.4)
             self.surfaceReceivedReset(applicationErrorCode: code)
+            #else
+
+            if self.pipelineStateMachine.isInitialized {
+                self.log("surfaceReset", metadata: ["applicationErrorCode": "\(code)"])
+                // yields buffered data and then surfaces the reset
+                self.streamRead()
+            } else {
+                // SM has captured the reset; the post-init `streamRead()`
+                // will surface it via `completeRead()`.
+                self.log("surfaceReset: pipeline not yet initialized, will surface on next read")
+            }
+            #endif
 
         case .doNothing(let reason):
             switch reason {
@@ -319,6 +332,7 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
         }
     }
 
+    #if compiler(<6.4)
     // Extracted from `handleReceivedResetStream` to keep that method's SIL
     // small enough to avoid the same `CopyPropagation` `-O` crash on
     // Swift 6.3.
@@ -336,6 +350,7 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
             self.log("surfaceReset: pipeline not yet initialized, will surface on next read")
         }
     }
+    #endif
 
     // Event that signals that STOP_SENDING was received
     internal func handleOutboundAbortedEvent(
--- a/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
+++ b/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
@@ -303,8 +303,8 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
                 promise: nil
             )
 
-        case .surfaceResetToInbound(let code):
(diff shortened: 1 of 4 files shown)


Sources/NIOQUIC/SwiftNetwork/QUICStreamStateMachine.swift
```diff
@@ -547,15 +568,20 @@ struct QUICStreamStateMachine: ~Copyable {
     }
 
     enum ReceiveResetStreamAction: ~Copyable {
-        /// Close the stream — both sides terminal.
-        case closeStream(WrappedStreamStateCriticalError)
-        /// Stream remains open (other side still active).
-        case doNotCloseStream(WrappedStreamStateCriticalError)
-        enum IgnoreReason: ~Copyable {
+        /// The receive side is the only remaining live direction —
```
josephnoir, line 550: Do you have a follow-up planned that removes the `WrappedStreamStateCriticalError` from `ReceiveStopSendingAction` as well?

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -296,24 +296,41 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
...
-            self.closeStream(
-                mode: .closeAndDisconnect,
-                error: NIOQUICHelpers.QUICStreamResetError(code: errorCode),
+                error: NIOQUICHelpers.QUICStreamResetError(code: code),
                 promise: nil
             )
 
-        case .ignore(.alreadyFullyReceived):
-            self.log("receiveResetStream: all data already received, ignoring")
+        case .surfaceResetToInbound(let code):
```
josephnoir: Nit: I don't like the `ToInbound` part of the name. `streamRead` is in the same type. Maybe `readStream` (matching `closeStream` above)?

josephnoir (commented):
LGTM overall. Just a nit and a question.

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
glbrntt: FWIW I find `readStream` problematic because it's unclear whether it's "the stream did read X" or "the stream should do a read".

Sources/NIOQUIC/SwiftNetwork/QUICStreamStateMachine.swift
```diff
@@ -547,15 +568,20 @@ struct QUICStreamStateMachine: ~Copyable {
...
+        /// tear down the channel with this error code.
+        case closeStream(applicationErrorCode: QUICApplicationErrorCode)
+        /// Surface the reset to the inbound pipeline.
+        case surfaceResetToInbound(applicationErrorCode: QUICApplicationErrorCode)
```
glbrntt: IMO these labels just add noise: the type fully encodes the label.

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
glbrntt: The naming also seems needlessly long to me: what's wrong with `deliver(QUICApplicationErrorCode)` or `deliverReset(QUICApplicationErrorCode)`?

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -296,24 +296,41 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
...
+                self.log("receiveResetStream: already fully received")
+            case .alreadyReset:
+                self.log("receiveResetStream: already reset")
+            }
+        }
+    }
+
+    // Extracted from `handleReceivedResetStream` to keep that method's SIL
+    // small enough to avoid the same `CopyPropagation` `-O` crash on
+    // Swift 6.3.
```
glbrntt, line 338: Is this resolved in 6.4 builds?

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -296,24 +296,41 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
...
+    ) {
+        if self.pipelineStateMachine.isInitialized {
+            self.log("surfaceResetToInbound: applicationErrorCode=\(code)")
+            // yields buffered data and then surfaces the reset
+            self.streamRead()
+        } else {
+            // SM has captured the reset; the post-init `streamRead()`
+            // will surface it via `completeRead()`.
+            self.log("surfaceResetToInbound: pipeline not yet initialized, will surface on next read")
         }
```
glbrntt, line 351: I assume the follow-up mentioned in the description will actually use the `code` that's passed in here?

Sources/NIOQUIC/SwiftNetwork/QUICStreamStateMachine.swift
```diff
@@ -259,6 +259,27 @@ struct QUICStreamStateMachine: ~Copyable {
...
+            }
+        case .pendingID:
+            return false
+        case .closed(let closed):
+            switch closed.reason {
+            case .peerReset: return true
+            case .clean, .localReset, .error: return false
+            }
+        }
+    }
```
glbrntt: This doesn't appear to be used?

Sources/NIOQUIC/SwiftNetwork/QUICStreamStateMachine.swift
rnro: I don't think so at this time; I think that would need to be a separate follow-up.

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
rnro: I have changed it to `surfaceReset`, what do you think of that? I don't like `streamRead` because I don't think it's right to be overly-prescriptive about _how_ it's surfaced, particularly because I'm a bit ambivalent about it happening via a read anyway.

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -657,8 +674,10 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
             }
             return true
 
-        case .reportPeerReset:
-            self.pipeline.fireUserInboundEventTriggered(ChannelEvent.inputClosed)
+        case .reportPeerReset(let code):
+            self.pipeline.fireErrorCaught(
+                NIOQUICHelpers.QUICStreamResetError(code: code)
```
glbrntt: You shouldn't need the `NIOQUICHelpers` prefix

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
josephnoir: The result is always an action. So while I don't think that `readStream` would be ambiguous, I am okay with using `deliver`/`surfaceReset`/`deliverReset` or similar.

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
rnro: It appears to be in my testing with nighties

Sources/NIOQUIC/SwiftNetwork/QUICStreamStateMachine.swift
rnro: Good spot

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
rnro: I think still just to be logged. However you have just made me realize that I should be using metadata rather than inline interpolation for all these logs according to best practices.

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
glbrntt: Can you add a dummy `#if compiler(<6.3)` / `#if compiler(>=6.4)` so that  we're more likely to find this later?

glbrntt (approved):

josephnoir (approved):

PR #44, opened 2026-07-13
Surface deferred peer `RESET_STREAM` at pipeline-init completion

rnro:
### Motivation

* When peer `RESET_STREAM` landed during the stream's pipeline-init window the reset stayed stashed in the SM and only surfaced on the next read; under `autoRead == false` an application that never reads never received `errorCaught(QUICStreamResetError)`.

### Modifications

* New `surfaceDeferredResetAfterInit` state holds the error and later yields it for the handler to fire.
* `QUICChannelStreamHandler` drains any stashed reset from each inbound `initialize` overload and from `createStreamChannel` once the pipeline is marked ready.

### Result

* The reset surfaces at init-complete regardless of whether reads were issued.


Sources/NIOQUIC/SwiftNetwork/QUICStreamStateMachine.swift
```diff
@@ -665,6 +694,34 @@ struct QUICStreamStateMachine: ~Copyable {
         }
     }
 
+    enum SurfaceDeferredResetAction: ~Copyable {
+        /// Caller should fire `errorCaught(QUICStreamResetError(code:))`.
+        case fireReset(applicationErrorCode: QUICApplicationErrorCode)
+        case nothing
```
josephnoir: Nit: We are using `doNothing` in other enums.

Sources/NIOQUIC/QUICConnectionChannelHandler.swift
```diff
@@ -451,15 +451,30 @@ extension QUICConnectionChannelHandler {
...
+                    case .success:
+                        switch streamHandler.pipelineStateMachine.markInitializerComplete() {
+                        case .surfaceInitializedStream:
+                            streamHandler.surfaceDeferredResetAfterInit()
+                            // Trigger the first read.
+                            streamHandler.tryToAutoRead()
+                        case .ignoreAlreadyComplete:
+                            break
+                        }
+                        channelActivationPromise.succeed(streamHandler)
```
josephnoir: Does the order matter here? When we succeed the promise after `surfaceDeferredResetAfterInit`, an error might be fired before anything attached to the promise runs.

rnro: It does! And my code actually had a bug 🙏. I've added a test.

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -493,6 +483,7 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
                     switch self.pipelineStateMachine.markInitializerComplete() {
                     case .surfaceInitializedStream:
                         continuation.yield(output: output, channel: self)
+                        self.surfaceDeferredResetAfterInit()
```
glbrntt, line 486: I'm not sure about this: couldn't there be buffered bytes which were received _before_ the reset that would now be delivered after? (Which would be very surprising for a user.)

glbrntt: You're also making some duplicated pipeline walks here: `surfaceDeferredResetAfterInit` calls `self.fireChannelReadComplete()` which also calls `self.tryToAutoRead()`. which we then call directly after this.

PR #47, opened 2026-07-14
Do not propagate stream channels on connection

josephnoir:
### Motivation

New streams were propagated by the connection channel handler. However, our adopter (SwiftNIO HTTP/3) never consumed these events. SwiftNIO QUIC already offers APIs for channel initialization that can be used for this purpose (see h3).

### Modifications

* Do not propagate new stream channels via the connection handler.
* Return the connection stream creator when creating outbound connections.
* Remove `HTTPConnectionHandler` from SyncIntergrationTests.

### Result

Slim down API a bit.


glbrntt (approved):

josephnoir:
Thank you for the review! API breakage check is expected to fail.

PR #49, opened 2026-07-16
Add connection, registrar, and transport interfaces

glbrntt:
Motivation:

The connection channel has to talk to a few different objects: the underlying QUIC connection, a connection ID registrar (for routing), and a transport (for writing outbound datagrams).

This PR introduces a few interfaces and wrapper types which currently only hold existential 'test' variants of each. In the fullness of time each variant will hold a concrete type of each as well to get static method dispatch for prod code while allowing enough flexibility for testing.

Modifications:

- Add protocols and enum wrappers for the registrar, connection, and transport
- Have the connection hold one of each

Result:

More pieces

### DIFF
--- a/Sources/NIOQUIC/QUICConnectionIDRegistrar.swift
+++ b/Sources/NIOQUIC/QUICConnectionIDRegistrar.swift
@@ -1,0 +1,64 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2026 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+@available(anyAppleOS 26, *)
+public protocol QUICConnectionIDRegistrar {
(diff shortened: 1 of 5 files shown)


Sources/NIOQUIC/QUICConnectionIDRegistrar.swift
```diff
@@ -0,0 +1,64 @@
...
+    /// Retires `connectionID`.
+    ///
+    /// - Parameter connectionID: The ID to retire.
+    /// - Returns: `false` if it wasn't registered.
+    func retire(_ connectionID: QUICConnectionID) -> Bool
+
+    /// Generates a new connection ID.
+    ///
+    /// - Returns: A new connection ID.
+    func generateID() -> QUICConnectionID
```
josephnoir, line 35: How does this interact with the [`QUICConnectionIDGenerator`](https://github.com/apple/swift-nio-quic/blob/main/Sources/NIOQUIC/QUICConnectionIDGenerator.swift#L44) that we currently have? It offers a second function that takes the CIDs chosen by the peer, which we should continue to support.

josephnoir (commented):
Looks good, just one question about ID generation.

Sources/NIOQUIC/QUICConnectionIDRegistrar.swift
glbrntt: It's just modeling the API currently used by `SwiftNetworkQUICConnection` -- this API will never be public so we can change it at will if we need to.

josephnoir (approved):

PR #50, opened 2026-07-22
Add lifecycle state machine to conn channel

glbrntt:
Motivation:

The connection channel needs to maintain a small amount of lifecycle state to know when to fire various pipeline operations and so on. This change adds a Lifecycle state machine which isn't yet wired up to the channel.

Modifications:

- Add a Lifecycle state machine

Result:

More things in place.

### DIFF
--- a/Sources/NIOQUIC/QUICConnectionChannel+Lifecycle.swift
+++ b/Sources/NIOQUIC/QUICConnectionChannel+Lifecycle.swift
@@ -220,10 +220,14 @@ extension QUICConnectionChannel {
             }
         }
 
-        /// Record that an inbound stream initializer finished. Decrements the gate; the
-        /// caller then runs `reconcileLifecycle()`, which fires any now-unblocked deferred
-        /// `channelInactive` via ``reconcile()``.
+        /// Record that an inbound stream initializer finished.
+        ///
+        /// Decrements the in-flight initializer count; this is used to hold up channel inactive
+        /// on the connection channel while streams are being initialized. At the end of the
+        /// outbound packet draining loop the channel should call ``reconcile()`` which may then
+        /// lead to the connection becoming inactive.
         mutating func streamInitializerFinished() {
+            assert(self.inFlightInitializers > 0)
             self.inFlightInitializers &-= 1
         }
 
@@ -236,10 +240,10 @@ extension QUICConnectionChannel {
             case fireInactive(error: (any Error)?)
         }
 
-        /// Reconcile the current lifecycle by applying a pending state.
+        /// Reconcile the current lifecycle by applying a pending state change.
         ///
-        /// The channel should call this in a loop until it returns `nil` when it has drained
-        /// its connection.
+        /// When the channel has finished consuming outbound data from the connection it should
+        /// call this in a loop and apply side effects until this returns `nil`.
         mutating func reconcile() -> Action? {
             switch self.state {
             case .initialized:
--- a/Sources/NIOQUIC/QUICConnectionChannel+Lifecycle.swift
+++ b/Sources/NIOQUIC/QUICConnectionChannel+Lifecycle.swift
@@ -1,0 +1,280 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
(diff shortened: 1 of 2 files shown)


Sources/NIOQUIC/QUICConnectionChannel+Lifecycle.swift
```diff
@@ -0,0 +1,280 @@
...
+            case .closed:
+                return false
+            }
+        }
+
+        /// Record that an inbound stream initializer finished. Decrements the gate; the
+        /// caller then runs `reconcileLifecycle()`, which fires any now-unblocked deferred
+        /// `channelInactive` via ``reconcile()``.
+        mutating func streamInitializerFinished() {
+            self.inFlightInitializers &-= 1
```
josephnoir, line 231: Can we add an `assert(self.inFlightInitializers > 0)` before decrementing it?

Sources/NIOQUIC/QUICConnectionChannel+Lifecycle.swift
```diff
@@ -0,0 +1,280 @@
...
+            /// A close was already in flight.
+            case alreadyClosing
+            /// The channel was already fully closed.
+            case alreadyClosed
+        }
+
+        /// Commit the channel to closing.
+        ///
+        /// The channel must later call ``reconcile()`` to get the side effects of
+        /// closing. The return value dictates whether the channel should proceed with
```
josephnoir, line 126: The docs around `reconcile` could be improved. Is the expectation that this is always called after every function that say it must be called later, is it only on specific return values, should it happen after every activation? This could go onto the `Lifecycle` type itself.

Sources/NIOQUIC/QUICConnectionChannel+Lifecycle.swift
```diff
@@ -0,0 +1,280 @@
...
+                return true
+            case .closing(let inactiveFired):
+                return !inactiveFired
+            case .closed:
+                return false
+            }
+        }
+
+        /// Record that an inbound stream initializer finished. Decrements the gate; the
+        /// caller then runs `reconcileLifecycle()`, which fires any now-unblocked deferred
```
josephnoir:
```suggestion
        /// caller then runs `reconcile()`, which fires any now-unblocked deferred
```

Sources/NIOQUIC/QUICConnectionChannel+Lifecycle.swift
```diff
@@ -0,0 +1,280 @@
...
+        enum Action {
+            /// Fire `channelActive`, complete the ready promise (if applicable), resume reads.
+            case fireActive
+            /// Fire `channelInactive` after firing the error (if present).
+            case fireInactive(error: (any Error)?)
+        }
+
+        /// Reconcile the current lifecycle by applying a pending state.
+        ///
+        /// The channel should call this in a loop until it returns `nil` when it has drained
```
josephnoir: Nit: "draining" is a [QUIC connection state](https://www.rfc-editor.org/rfc/rfc9000.html#section-10.2.2), but I assume that is not what "drained" refers to here.

glbrntt: No, it's referring to the "drain" terminology used by `SwiftNetworkQUICConnection`, i.e. pulling outbound data out so it can be written to the socket. I updated the docs here.

Sources/NIOQUIC/QUICConnectionChannel+Lifecycle.swift
glbrntt: Yes!

Sources/NIOQUIC/QUICConnectionChannel+Lifecycle.swift
glbrntt: That's fair, I think it's a little hard to follow without the channel using this. I'll update the docs. The channel consumes packets from `SwiftNetworkQUICConnection` in a loop, during that loop various other events may happen which trigger the pending state. `reconcile` is then called at the end of the loop to apply those changes.

josephnoir (approved):
LGTM. Thanks, George!

PR #51, opened 2026-07-22
Add support for QUIC datagrams

josephnoir:
### Motivation

QUIC datagrams extends QUIC with unreliable datagrams, defined in RFC 9221. SwiftQUIC support them and SwiftNIO QUIC should too.

### Modifications

Add a new channel handler that sits in the connection channel and sends and receives QUIC datagrams via SwiftQUIC. Writes in the channel turn into outgoing datagrams while reads deliver incoming datagrams.

**Note:** The current implementation does not propatate the peer's advertised limit and always assumes a 65k frame size. This will be updated in a follow-up.

### Results

Support for sending and receiving QUIC datagrams.


josephnoir:
Closing in favor or smaller PRs.

PR #52, opened 2026-07-22
Add QUIC datagram CH with a test backend

josephnoir:
### Motivation

RFC 9221 extends QUIC with unreliable datagrams. SwiftQUIC supports them and SwiftNIO QUIC should too. This adds a channel handler, but defers the implementation of backend.

### Modifications

Add `QUICDatagramHandler`, a channel handler to send and receive QUIC datagrams with a configurable backend. The only supported backend at the moment is a test backend.

The channel handler can be installed before the connection is fully established as it does not forward datagrams to the backend until the connection was established and the frame size advertised by the peer is known.

### Results

Prepare QUIC datagram support for SwiftNIO QUIC.

### DIFF
--- a/Sources/NIOQUIC/QUICDatagramHandler.swift
+++ b/Sources/NIOQUIC/QUICDatagramHandler.swift
@@ -70,9 +70,6 @@ final class QUICDatagramHandler: ChannelDuplexHandler {
         transport.setReader(reader: self)
         self.transport = .test(transport)
         self.setPeerMaxDatagramFrameSize(size)
-        if let context = self.context {
-            self.flush(context: context)
-        }
     }
 }
 
@@ -256,6 +253,9 @@ extension QUICDatagramHandler {
                     } else {
                         promise?.fail(QUICError.datagramWriteFailed)
                     }
+                }
+                if earlyWrites.count > 0 {
+                    self.transport.flush()
                 }
             }
         case .peerDoesNotAcceptDatagrams:
--- a/Sources/NIOQUIC/QUICDatagramHandler.swift
+++ b/Sources/NIOQUIC/QUICDatagramHandler.swift
@@ -1,0 +1,281 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2026 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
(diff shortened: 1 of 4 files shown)


Sources/NIOQUIC/QUICDatagramHandler.swift
```diff
@@ -0,0 +1,281 @@
...
+    /// Installs a `QUICDatagramProtocol` conforming backend for testing, applying the peer's
+    /// advertised `max_datagram_frame_size` the same way `setBackend(to:withPeerMaxDatagramFrameSize:)`
+    /// does. Writes buffered before this call are resolved against `size`.
+    func setTestBackend(to transport: any QUICDatagramProtocol, peerMaxDatagramFrameSize size: Int) {
+        transport.setReader(reader: self)
+        self.transport = .test(transport)
+        self.setPeerMaxDatagramFrameSize(size)
+        if let context = self.context {
+            self.flush(context: context)
+        }
```
glbrntt: why the unconditional flush?

Sources/NIOQUIC/QUICDatagramProtocol.swift
```diff
@@ -0,0 +1,58 @@
...
+    ///   surfaces to the caller as `QUICError.datagramWriteFailed`.
+    func write(datagram: ByteBuffer) -> Bool
+
+    /// Sends any datagrams buffered since the last `flush()`.
+    func flush()
+
+    /// Detaches the underlying flow and clears the current reader.
+    func close()
+
+    /// Sets the delegate that receives datagrams and errors from this transport.
```
glbrntt, line 42: Are you sure this should be part of the protocol? Ideally the "real" type will hold the concrete reader. Although I guess the "real" type can have an overload and store the concrete type anyway.

glbrntt (commented):
Looks good modulo a couple of questions.

Sources/NIOQUIC/QUICDatagramHandler.swift
josephnoir: Good point. I'll make it conditional on the early writes.

Sources/NIOQUIC/QUICDatagramProtocol.swift
josephnoir: The overload was my intention. ~~Although that would require a return type to propagate if the concrete type does not accept the generic one.~~ Edit: That type will get an enum.

glbrntt (approved):

PR #53, opened 2026-07-23
Add SwiftNetwork backend for QUIC Datagrams

josephnoir:
### Motivation

Without a real backend QUIC datagrams remain unusable.

### Modifications

* Implement a SwiftNetwork-based backend for QUIC datagrams.
* Add a new configuration option to configure the advertised max datagram frame size.

### Results

QUIC datagrams can be sent and received via the QUIC connection channel.

#### Note

The peer's advertised max frame size is not yet available and estimated to be the RFC recommended value.

### DIFF
--- a/Sources/NIOQUIC/QUICDatagramHandler.swift
+++ b/Sources/NIOQUIC/QUICDatagramHandler.swift
@@ -65,10 +65,22 @@ final class QUICDatagramHandler: ChannelDuplexHandler {
 
     /// Installs a `QUICDatagramProtocol` conforming backend for testing, applying the peer's
     /// advertised `max_datagram_frame_size` the same way `setBackend(to:withPeerMaxDatagramFrameSize:)`
-    /// does. Writes buffered before this call are resolved against `size`.
+    /// does.
     func setTestBackend(to transport: any QUICDatagramProtocol, peerMaxDatagramFrameSize size: Int) {
         transport.setReader(reader: self)
         self.transport = .test(transport)
+        self.setPeerMaxDatagramFrameSize(size)
+    }
+
+    /// Installs a SwiftNetwork-backed transport.
+    ///
+    /// - Parameters:
+    ///   - transport: The transport to send and receive datagrams through.
+    ///   - size: The peer's advertised `max_datagram_frame_size`. `0` means the peer does not
+    ///     accept datagrams. Must be >= 0. Buffered and new packets are verified to stay below this limit.
+    func setBackend(to transport: QUICDatagramTransport, withPeerMaxDatagramFrameSize size: Int) {
+        transport.setReader(reader: self)
+        self.transport = .swiftNetwork(transport)
         self.setPeerMaxDatagramFrameSize(size)
     }
 }
@@ -84,6 +96,8 @@ extension QUICDatagramHandler {
         case none
         /// Installed via `setTestBackend(to:withPeerMaxDatagramFrameSize:)`.
         case test(any QUICDatagramProtocol)
+        /// Installed via `setBackend(to:withPeerMaxDatagramFrameSize:)`.
+        case swiftNetwork(QUICDatagramTransport)
     }
 }
 
@@ -100,6 +114,8 @@ extension QUICDatagramHandler.Transport: QUICDatagramProtocol {
             fatalError("state violation: cannot perform a write before assigning a transport")
         case .test(let testTransport):
             testTransport.write(datagram: datagram)
+        case .swiftNetwork(let swiftNetworkTransport):
+            swiftNetworkTransport.write(datagram: datagram)
         }
     }
 
@@ -110,6 +126,8 @@ extension QUICDatagramHandler.Transport: QUICDatagramProtocol {
             break
         case .test(let testTransport):
             testTransport.flush()
+        case .swiftNetwork(let swiftNetworkTransport):
+            swiftNetworkTransport.flush()
         }
     }
 
@@ -120,6 +138,8 @@ extension QUICDatagramHandler.Transport: QUICDatagramProtocol {
             break
         case .test(let testTransport):
             testTransport.close()
+        case .swiftNetwork(let swiftNetworkTransport):
+            swiftNetworkTransport.close()
         }
     }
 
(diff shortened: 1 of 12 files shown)


josephnoir:
The API breakage check hits the new argument in the `QUICConfiguration` initializers. The messages sent on the connection channel also changed, which could break adopters as well.

Sources/NIOQUIC/QUICDatagramHandler.swift
```diff
@@ -65,12 +65,24 @@ final class QUICDatagramHandler: ChannelDuplexHandler {
...
+
+    /// Installs a SwiftNetwork-backed transport.
+    ///
+    /// - Parameters:
+    ///   - transport: The transport to send and receive datagrams through.
+    ///   - size: The peer's advertised `max_datagram_frame_size`. `0` means the peer does not
+    ///     accept datagrams. Must be >= 0. Buffered and new packets are verified to stay below this limit.
+    func setBackend(to transport: QUICDatagramTransport, withPeerMaxDatagramFrameSize size: Int) {
+        transport.setReader(reader: self)
+        self.transport = .swiftNetwork(transport)
```
glbrntt, line 83: When do we break this retain cycle? In `handlerRemoved` we only call `close()`. Presumably we should unset it there as well.

Sources/NIOQUIC/QUICDatagramHandler.swift
```diff
@@ -132,6 +152,8 @@ extension QUICDatagramHandler.Transport: QUICDatagramProtocol {
             fatalError("state violation: cannot set a reader before assigning a transport")
         case .test(let testTransport):
             testTransport.setReader(reader: reader)
+        case .swiftNetwork(let swiftNetworkTransport):
+            swiftNetworkTransport.setReader(reader: reader)
```
glbrntt: Should've realized this in the previous PR, but `reader:` is needless here: the argument to pass to `setReader` is unambiguously a reader so the call site should be: `setReader(reader)`. See also "Omit needless words" from https://www.swift.org/documentation/api-design-guidelines/#promote-clear-usage

Sources/NIOQUIC/SwiftNetwork/SwiftNetworkQUICConnection.swift
```diff
@@ -95,6 +95,8 @@ final class SwiftNetworkQUICConnection {
     private var streamInputHandlers: [QUICStreamID: QUICChannelStreamHandler] = [:]
     private var pendingInitialClientStream: QUICChannelStreamHandler?
 
+    private var datagramHandler: QUICDatagramHandler?
```
glbrntt: We should break the retain cycle in here as well.

Sources/NIOQUIC/SwiftNetwork/QUICConfiguration.swift
```diff
@@ -111,6 +111,9 @@ public struct QUICConfiguration: Sendable {
     public var qLogConfiguration: QLogConfiguration?
     /// Configure verification of the peer's certificate. Only supported by the client.
     public var peerCertificateVerification: CertificateVerification
+    /// Maximum datagram frame size in bytes. Set to 0 to disable datagrams.
+    /// Defaults to 65535 (the max value) as recommended in RFC 9221.
```
glbrntt:
```suggestion
    /// Defaults to 65535 (the max value allowed) as recommended in RFC 9221.
```

Sources/NIOQUIC/SwiftNetwork/QUICChannelNewFlowHandler.swift
```diff
@@ -302,3 +311,64 @@ extension QUICChannelNewFlowHandler: UpperProtocolHandler {
...
+                transport.reference,
+                remote: nil,
+                local: nil,
+                parameters: swiftNetworkParameters,
+                path: swiftNetworkPath
+            )
+            transport.setFlowLinkage(linkage)
+            linkage.invokeConnect(transport.reference)
+
+            // TODO: Once SwiftNetwork (a40d5771fc586e75f3534e55178398baa1db2966) is released,
```
glbrntt, line 361: Is there anything blocking a release, if not we should just ask them to do it. FWIW we should do some releases too!

Sources/NIOQUIC/SwiftNetwork/QUICStreamOptions+QUICConfiguration.swift
```diff
@@ -33,6 +33,7 @@ extension QUICProtocol {
         options.connectionOptions.maximumConcurrentBidirectionalStreams = config.initialMaxStreamsBidi
         options.connectionOptions.initialMaxStreamsUnidirectional = UInt64(config.initialMaxStreamsUni)
         options.connectionOptions.maximumConcurrentUnidirectionalStreams = config.initialMaxStreamsUni
+        options.connectionOptions.maxDatagramFrameSize = UInt16(truncatingIfNeeded: config.maxDatagramFrameSize)
```
glbrntt: I don't think this does what you think it does 🙂:

```swift
UInt16(truncatingIfNeeded: Int(UInt16.max) + 1) == 0
```

glbrntt (changes requested):
Couple of small things but this largely looks good!

Sources/NIOQUIC/QUICDatagramHandler.swift
josephnoir: Definitely!

Sources/NIOQUIC/QUICDatagramHandler.swift
josephnoir: `close()` will remove the ref in the transport, so there shouldn't be a cycle. But we should clean this up here as well. It's not usable anymore.

Sources/NIOQUIC/SwiftNetwork/SwiftNetworkQUICConnection.swift
josephnoir: Good spot! This is not used though, so it shouldn't be here at all. It's a leftover from my initial approach.

Sources/NIOQUIC/SwiftNetwork/QUICStreamOptions+QUICConfiguration.swift
josephnoir: No, it is not. 😂

Sources/NIOQUIC/SwiftNetwork/QUICChannelNewFlowHandler.swift
josephnoir: I don't think there is.

josephnoir:
Thank you for the review!

glbrntt (approved):

PR #54, opened 2026-07-24
Wire up connection channel lifecycle and I/O

glbrntt:
Motivation:

Much of the bones of the connection channel has landed, most of it just needs wiring up.

Modifications:

- Add connection and transport views; these are the interfaces used by SwiftNetworkQUICConnection and the parent UDP channel.
- Wire up lifecycle
- Some API (streams) isn't testable as it relies on SwiftNetwork

Result:

Channel is functional with test mocks

### DIFF
--- a/Sources/NIOQUIC/QUICConnectionChannel.swift
+++ b/Sources/NIOQUIC/QUICConnectionChannel.swift
@@ -69,8 +69,24 @@ final class QUICConnectionChannel: @unchecked Sendable {
     /// A view into the QUIC handler in the UDP channel, for outbound operations.
     private let _transport: Transport
 
+    /// The lifecycle state machine for the channel.
+    private var _lifecycle: Lifecycle
+
+    /// Initializer for inbound streams.
+    private var _streamInitializer: StreamInitializer?
+
+    /// Promise to complete when the channel becomes active (or closed if never active).
+    private var _readyPromise: EventLoopPromise<Void>?
+
     /// Whether auto-read is enabled on this channel.
     private var _autoRead: Bool
+
+    /// Whether the channel is allowed to drain the output from the connection.
+    private var _isAllowedToDrain: Bool
+
+    /// A queue of streams pushed to the channel by the underlying connection. These are dequeued
+    /// at the end of the parent channel's read loop (in `_parentChannelReadComplete`).
+    private var _pendingStreams: Deque<(QUICStreamID, QUICChannelStreamHandler)>
 
     // MARK: - Channel API
 
@@ -105,9 +121,22 @@ final class QUICConnectionChannel: @unchecked Sendable {
         self._connection = connection
         self._registrar = registrar
         self._transport = transport
+        self._lifecycle = Lifecycle()
         self._autoRead = true
+        self._streamInitializer = nil
+        self._readyPromise = nil
+        self._isAllowedToDrain = true
+        self._pendingStreams = []
 
         self._pipeline = ChannelPipeline(channel: self)
+    }
+
+    enum StreamInitializer {
+        /// Hand new streams to a multiplexer continuation. Used by the typed-output
+        /// `QUICConnection<Output>` path.
+        case multiplexer(any StreamMultiplexerContinuation)
+        /// Initialize new streams via the supplied closure.
+        case closure(@Sendable (any Channel) -> EventLoopFuture<Void>)
     }
 }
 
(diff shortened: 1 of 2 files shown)


Sources/NIOQUIC/QUICConnectionChannel.swift
```diff
@@ -275,3 +316,634 @@ extension QUICConnectionChannel: ChannelCore {
...
+    /// frames, or `close()` stopping the flow handler).
+    ///
+    /// The explicit drain after flushes whatever `body` finalized. Without this
+    /// the re-entrant wake would run a nested `drainAndApplyLifecycle()`
+    /// mid-batch, firing `channelActive`/`Inactive` before the batch completes.
+    private func withoutEnteringDrainOutput<Result>(_ body: () -> Result) -> Result {
+        self._isAllowedToDrain = false
+        defer { self._isAllowedToDrain = true }
+        return body()
+    }
```
josephnoir, line 645: Not sure if we can run into this here, but if you call this is called while `self._isAllowedToDrain` is already `false` this will still flip it to `true`.

As an alternative to restoring the previous state, this could `guard self._isAllowedToDrain else { return }` at the beginning, or have an `assert(self._isAllowedToDrain == true)`.

Sources/NIOQUIC/QUICConnectionChannel.swift
```diff
@@ -275,3 +316,634 @@ extension QUICConnectionChannel: ChannelCore {
...
+            shouldClose = true
+        case .alreadyClosing, .alreadyClosed:
+            shouldClose = false
+        }
+
+        guard shouldClose else { return }
+
+        self.drainOutput()
+
+        // TODO: what's the rationale here? Should we close and then drain?
```
josephnoir: 👀

Sources/NIOQUIC/QUICConnectionChannel.swift
```diff
@@ -275,3 +316,634 @@ extension QUICConnectionChannel: ChannelCore {
...
+
+        let shouldClose: Bool
+        switch self._lifecycle.beginClosing(error: nil) {
+        case .beganClosing:
+            shouldClose = true
+        case .alreadyClosing, .alreadyClosed:
+            shouldClose = false
+        }
+
+        guard shouldClose else { return }
```
josephnoir, line 755: Nit: In other cases we one case `return`s while the passing case has `()`. See `closeConnection` above. Any reason not to do the same here?

Sources/NIOQUIC/QUICConnectionChannel.swift
```diff
@@ -275,3 +316,634 @@ extension QUICConnectionChannel: ChannelCore {
...
+        self._pendingStreams.append((streamID, channel))
+    }
+
+    /// Initialize inbound streams pushed during the current read batch, skipping any not yet
+    /// connected. Returns `true` if any stream was initialized.
+    private func processPendingInboundStreams() -> Bool {
+        self.eventLoop.assertInEventLoop()
+        var anyStreamInitialized = false
+
+        while let (streamID, stream) = self._pendingStreams.popFirst() {
```
josephnoir, line 909: Just to clarify: Since these are popped, are they re-added later when they fail the check in line 910? I only see the one append in `_newInboundStream(streamID:channel:)` above.

josephnoir (changes requested):
Looking good overall!

Sources/NIOQUIC/QUICConnectionChannel.swift
glbrntt: It matches the existing behavior on `main`: https://github.com/apple/swift-nio-quic/blob/6fedbc13754d3446188dea308de7660577e14f68/Sources/NIOQUIC/SwiftNetwork/SwiftNetworkQUICConnection.swift#L959-L974 

I _think_ it makes sense: `isConnected` returns true if `.connected` (...duh) and `false` for `.pendingID`/`.closed`. 

Because this is an inbound stream it'll never be in `.pendingID`, and if it's already closed (weird but possible, maybe? 🤷‍♂️) then we shouldn't initialize it.

Sources/NIOQUIC/QUICConnectionChannel.swift
glbrntt: I tend to favor minimizing the number of `return`s where possible so I'd rather  fix the one above 😄 -- don't mind either way though, what would you prefer?

josephnoir: I care more about recognizable patterns than the exact spelling. You are welcome to pick the one you prefer. :)

Sources/NIOQUIC/QUICConnectionChannel.swift
josephnoir: I see, that makes sense. Thanks!

Sources/NIOQUIC/QUICConnectionChannel.swift
glbrntt: Yeah, good, catch we can reach it.

josephnoir (approved):
Thanks!

PR #56, opened 2026-07-28
Wire up connection channel

glbrntt:
Motivation:

A bunch of changes have been paving the way to make this possible. Now it's a case of putting the pieces together. There will be a follow up after this change to delete a bunch of unused code. For now this change only deletes code that no longer compiles.

Modifications:

- Replace ChildChannelMultiplexer in the QUICHandler with a registry of connection channel transport views, keyed by connection ID
- Wire up the "live" variants of the objects that the connection channel holds
- Move datagrams from being a channel handler to being on the connection channel
- Delete QUICConnectionChannelHandler, it's superseded by the new connection channel
- Drop the channel-shaped callbacks from SwiftNetworkQUICConnection in favour of a single channel view
- Remove 'QUICMetrics.Connection.openStreams' and 'QUICDrainOutputEvent', which no longer have anything to record or drive.

Result:

Fewer moving parts, more perf.


Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -472,40 +498,78 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
...
+                }
+        case .ignore(let reason):
+            switch consume reason {
+            case .initializerComplete:
+                promise.succeed(self)
+                self.tryToAutoRead()
+            case .channelInactive:
+                promise.succeed(self)
+            case .initializerInProgress:
+                promise.succeed(self)
```
josephnoir, line 582: The docs say "[...] succeeding `promise` with the stream once it is ready.", but the last two cases don't sounds like that is the case. Is this intended?

Sources/NIOQUIC/SwiftNetwork/QUICConnectionStateMachine.swift
```diff
@@ -622,37 +622,39 @@ struct QUICConnectionStateMachine: ~Copyable {
 
     enum OutOfBandWriteRequestAction: ~Copyable {
         case unexpectedRequest
-        case triggerEvent(on: any Channel)
+        case triggerEvent(QUICConnectionChannel.ConnectionView)
```
josephnoir: (Very much a nit, but I liked the `on:` label.)

Sources/NIOQUIC/SwiftNetwork/SwiftNetworkQUICConnection.swift
```diff
@@ -867,10 +819,26 @@ final class SwiftNetworkQUICConnection {
     /// Cleans up callbacks to break retain cycles before connection teardown
     private func cleanupCallbacks() {
         self.log("Cleaning up connection callbacks")
-        // This ensures the closure that captures QUICHandler is released
-        self.associateConnectionID = nil
-        self.retireConnectionID = nil
-        self.generateConnectionID = nil
+        self.channelView = nil
+    }
```
josephnoir: Do we still need this or can we use the new `dropChannelReferences` below instead?

Sources/NIOQUIC/QUICHandler.swift
```diff
@@ -802,212 +715,86 @@ extension QUICHandler: ChannelInboundHandler {
...
+                switch result {
+                case .success:
+                    self.deliverFirstPacket(addressedEnvelope.data, to: view)
+                case .failure:
+                    self.connectionRegistry.removeValue(forKey: view.registeredConnectionID)
                 }
             }
 
-            // We keep track of the original DCID the peer used in case they retransmit or send additional INITIAL packets.
-            // TODO: Remmove dcid from multiplexing.
```
josephnoir, line 869: I don't think this was addressed. We should keep it around here to create an issue.

josephnoir (commented):
It's nice to see all the previous PRs come together! LGTM. I only have a few minor things.

josephnoir (approved):
Thanks!

PR #57, opened 2026-07-29
Add a coalescer for GSO

glbrntt:
Motivation:

We should support GSO. To do so we'll need to coalesce outbound datagrams to the same remote address under certain conditions: each datagram in a run must be the same size (athough the last may be shorter), a run must not exceed the max length (typically 64) and a coalesced buffer must not exceed a max size (typically 64KiB).

This change adds a coalescer to do just that. It doesn't yet wire it up though. That'll be added in a separate PR.

Modifications:

- Add a coalescer for GSO

Result:

Nothing, it isn't used yet.


Sources/NIOQUIC/GSOCoalescer.swift
```diff
@@ -0,0 +1,117 @@
...
+import NIOCore
+
+/// Groups an ordered burst of outbound datagrams for one peer into runs which can be sent to the
+/// kernel as a single UDP Generic Segmentation Offload write.
+struct GSOCoalescer {
+    /// The maximum number of segments to coalesce.
+    let maxSegments: Int
+
+    /// The maximum size of a coalesced datagram.
+    let maxCoalescedSize: Int
```
josephnoir:
```suggestion
    var maxSegments: Int

    /// The maximum size of a coalesced datagram.
    var maxCoalescedSize: Int
```
Should these be `var`s?

Sources/NIOQUIC/GSOCoalescer.swift
```diff
@@ -0,0 +1,117 @@
...
+            // Bigger; end run without including this buffer.
+            if size > segmentSize { break }
+
+            totalSize &+= size
+            runLength &+= 1
+
+            // Smaller; end run the run.
+            if size < segmentSize { break }
+
+            // Same size: continue iterating.
```
josephnoir:
```suggestion
            // Same size; continue iterating.
```

Sources/NIOQUIC/GSOCoalescer.swift
```diff
@@ -0,0 +1,117 @@
...
+            self.buffers.formIndex(after: &index)
+        }
+
+        if runLength == 1 {
+            // Nothing to coalesce; so just remove and return
```
josephnoir:
```suggestion
            // Nothing to coalesce; so just remove and return.
```

josephnoir (approved):
Looks good to me!

PR #58, opened 2026-07-29
Use a stable ID for connections

glbrntt:
Motivaiton:

The recent connection channel changes rely on connection IDs for routing which makes tracking state a little tricky as new IDs can be associated and old IDs retired. We can simplify this by assigning each connection a unique in-process ID instead and using that as the canonical ID for a given connection.

Modifications:

- Add a ConnectionRegistry and drop AliasDictionary
- Write up ConnectionRegistry in the QUICHandler

Result:

Simpler


Sources/NIOQUIC/QUICHandler.swift
```diff
@@ -943,15 +945,37 @@ extension QUICHandler.RegistrarView: Sendable {}
...
             transport: .live(QUICHandler.ChildView(self)),
-            isServer: quicConnection.role == .server,
-            sourceConnectionID: sourceConnectionID
+            isServer: quicConnection.role == .server
         )
     }
 }
+
+/// An opaque, stable identity for a QUIC connection.
+struct ConnectionHandle: Hashable, Sendable {
```
josephnoir, line 961: We might be able to reuse `OpaqueIDGenerator` here, but it's probably good to have an explicit type.

josephnoir (approved):
I approve, thanks! :)

Sources/NIOQUIC/QUICHandler.swift
glbrntt: Yeah, I much prefer explicit types for this kind of thing. It's also not very opaque if the underlying type is part of the signature 😉

glbrntt: FWIW the explicit types are much better for logging/debugging too as they can include a snippet of what they're an ID for in their description, see here for prior art: https://github.com/grpc/grpc-swift-nio-transport/blob/c46f77c07c3ae4fce4734e7af03fcb35ab544428/Sources/GRPCNIOTransportCore/Internal/ProcessUniqueID.swift#L37-L61

PR #60, opened 2026-07-29
Delete unused code and unwired metrics

glbrntt:
Motivation:

Unused code is bad!

Modifications:

- Remove 'QUICMetrics', 'SwiftNetworkQUICConnection.Metrics', 'currentMetrics()', every 'metrics:' parameter, and the swift-metrics dependency. It wasn't populated and is dead weight. We already have an issue tracking support for this but we should think it through fully when we work on it.
- Delete various funcs/members with no callers
- Drop parameters every conformer ignores, e.g. 'channel:' from 'yield(connection:)'.

Result:

- Less code
- Fewer dependencies


josephnoir (approved):
Nice. Can we quickly confirm with @rnro that `detachAndClose` can be removed and is not required in an upcoming change.

rnro:
👍 I think I got to the end of my refactoring work, I can re-add it easily enough if it does turn out to be needed.

PR #61, opened 2026-07-31
Hold outbound writes while in a read loop

glbrntt:
Motivation:

The QUIC connection should only proactively inform the connection channel of outbound writes when the write occurs out-of-band (i.e. not during a read loop).

Currently that mechanism is missing which means every write is an out-of-band write, leading to more flushes than is necessary.

Modifications:

- Add an 'inReadLoop' flag to the quic connection and use it to hold out-of-band writes
- The connection channel polls for outbound writes at the end of the loop rather than being notified by the connection that it should consume the writes immediately

Result:

More perf

### DIFF
--- a/Sources/NIOQUIC/SwiftNetwork/SwiftNetworkQUICConnection.swift
+++ b/Sources/NIOQUIC/SwiftNetwork/SwiftNetworkQUICConnection.swift
@@ -83,6 +83,17 @@ final class SwiftNetworkQUICConnection {
     /// finalizes frames outside any drain bracket initiated by the channel.
     internal var channelView: QUICConnectionChannel.ConnectionView?
 
+    /// Whether the connection is in a read-loop.
+    ///
+    /// While this is `true`, outbound writes should not be emitted. Entering the read loop is
+    /// implicit from a call to `receivePacket()`, exiting the loop is explicit via a call to
+    /// `exitReadLoop()`.
+    private var inReadLoop: Bool
+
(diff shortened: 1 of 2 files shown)


Sources/NIOQUIC/SwiftNetwork/SwiftNetworkQUICConnection.swift
```diff
@@ -360,6 +370,7 @@ final class SwiftNetworkQUICConnection {
             context: swiftNetworkParameters.context
         )
 
+        self.inReadLoop = false
```
fabianfett, line 373: it's a tiny bit weird, that we have a method for this, that we don't invoke here. take it or leave it.

glbrntt: This is the `init`

fabianfett: right. safe that line and just initialize at the declaration site?

glbrntt: I much prefer doing all the initialization in `init` otherwise you can end up in split brain world where the declaration does one thing and the init does something else which doesn't agree. Sometimes you also need to init something and then do extra work (like reserve capacity). Doing it all in `init` means it's always consistent.

rnro (approved):

PR #62, opened 2026-07-31
Uses Frame as input for GSO coalescing

glbrntt:
Motivation:

In 05cf814 a `ByteBuffer` based coalescer was added for GSO. It occurred to me afterwards that there were two more optimisation that could be made.

The first is to consume `Frame` directly rather than going via an intermediate `ByteBuffer`. The other is reusing a pool of byte buffers rather than coalescing into the first buffer. As coalescing is done in the read loop there should be relatively few buffers used (if coalescing is successful) meaning we can reuse them between read cycles and save the allocation costs.

Modifications:

- Add `BufferPool`, a stripped down and simplified version of NIOs `NIOPooledRecvBufferAllocator`.
- Change the coalescer to store frames

Result:

GSO will be cheaper when we wire it up

### DIFF
--- a/Sources/NIOQUIC/GSOCoalescer.swift
+++ b/Sources/NIOQUIC/GSOCoalescer.swift
@@ -13,55 +13,105 @@
 //===----------------------------------------------------------------------===//
 
 import DequeModule
-import NIOCore
+@_spi(CustomByteBufferAllocator) import NIOCore
+@_spi(Essentials) @_spi(ProtocolProvider) import SwiftNetwork
 
 /// Groups an ordered burst of outbound datagrams for one peer into runs which can be sent to the
 /// kernel as a single UDP Generic Segmentation Offload write.
-struct GSOCoalescer {
+@available(anyAppleOS 26, *)
+struct GSOCoalescer: ~Copyable {
     /// The maximum number of segments to coalesce.
     var maxSegments: Int
 
     /// The maximum size of a coalesced datagram.
     var maxCoalescedSize: Int
 
-    private var buffers: Deque<ByteBuffer>
+    /// Frames to coalesce.
+    private var frames: UniqueDeque<Frame>
+
+    /// The address of the remote peer to send datagrams to.
     private let remoteAddress: SocketAddress
 
-    init(remoteAddress: SocketAddress, maxSegments: Int = 64, maxCoalescedSize: Int = 65535) {
+    /// A pool of byte buffers.
+    private var pool: BufferPool
+
+    init(
+        remoteAddress: SocketAddress,
+        maxSegments: Int = 64,
+        maxCoalescedSize: Int = 65535,
+        bufferPoolCapacity: Int = 8
+    ) {
         self.remoteAddress = remoteAddress
         self.maxSegments = maxSegments
         self.maxCoalescedSize = maxCoalescedSize
-        self.buffers = Deque()
-        self.buffers.reserveCapacity(16)
+
+        self.frames = UniqueDeque<Frame>()
+        self.frames.reserveCapacity(16)
+
+        self.pool = BufferPool(capacity: bufferPoolCapacity, allocator: ByteBufferAllocator())
     }
 
     var isEmpty: Bool {
-        self.buffers.isEmpty
+        self.frames.isEmpty
     }
 
-    mutating func append(_ buffer: ByteBuffer) {
-        if buffer.readableBytes > 0 {
-            self.buffers.append(buffer)
+    mutating func finalizeAllFramesAsFailed() {
+        while var frame = self.frames.popFirst() {
+            frame.finalize(success: false)
         }
+    }
+
+    mutating func append(frames: consuming FrameArray) {
+        self.frames.reserveCapacity(self.frames.count + frames.count)
+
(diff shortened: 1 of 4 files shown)


Sources/NIOQUIC/GSOCoalescer.swift
```diff
@@ -13,55 +13,105 @@
...
+        bufferPoolCapacity: Int = 8
+    ) {
         self.remoteAddress = remoteAddress
         self.maxSegments = maxSegments
         self.maxCoalescedSize = maxCoalescedSize
-        self.buffers = Deque()
-        self.buffers.reserveCapacity(16)
+
+        self.frames = UniqueDeque<Frame>()
+        self.frames.reserveCapacity(16)
```
rnro, line 49: Do you think there's any merit in this being configurable? It seems to be the only remaining magic number.

Sources/NIOQUIC/GSOCoalescer.swift
```diff
@@ -13,55 +13,105 @@
...
+        while var frame = self.frames.popFirst() {
+            frame.finalize(success: false)
+        }
+    }
+
+    mutating func append(frames: consuming FrameArray) {
+        self.frames.reserveCapacity(self.frames.count + frames.count)
+
+        while var frame = frames.popFirst() {
+            if frame.unclaimedLength == 0 {
```
rnro, line 68: In this case we have reserved capacity for one or more frames which are never appended, do you know under what conditions this can happen or if it's likely? I'm wondering if it is worth the cost of deciding how many frames actually need to be appended when we reserve the capacity.

Sources/NIOQUIC/GSOCoalescer.swift
glbrntt: No, not really. It's just to skip the first few reallocations.

Sources/NIOQUIC/GSOCoalescer.swift
glbrntt: I think it'd be very very unlikely: we're consuming a `FrameArray` of outbound datagrams produced by SwiftQUIC, it _shouldn't_ ever produce an empty `Frame` but it'd be good to protect against as it would break a GSO run and is otherwise just noise.

Sources/NIOQUIC/GSOCoalescer.swift
```diff
@@ -13,55 +13,105 @@
...
+        }
     }
 
-    mutating func append(_ buffer: ByteBuffer) {
-        if buffer.readableBytes > 0 {
-            self.buffers.append(buffer)
+    private mutating func datagram(from frame: consuming Frame) -> AddressedEnvelope<ByteBuffer> {
+        let buffer: ByteBuffer
+
+        if let config = frame.takeOwnershipOfCustomFinalizerBuffer() {
```
rnro: nit: is `config` the most appropriate name here? Something like `frameInternals` makes more sense to me.

Sources/NIOQUIC/GSOCoalescer.swift
```diff
@@ -13,55 +13,105 @@
...
+            )
+        } else {
+            var buf = ByteBuffer()
+            buf.reserveCapacity(frame.unclaimedLength)
+            frame.span?.withUnsafeBufferPointer { _ = buf.writeBytes($0) }
+            frame.finalize(success: true)
+            buffer = buf
         }
+
+        return AddressedEnvelope(remoteAddress: self.remoteAddress, data: buffer)
```
rnro, line 95: Should this have metadata if the one at the bottom of `next` does?

Sources/NIOQUIC/GSOCoalescer.swift
glbrntt: Probably not, I copied it from elsewhere and didn't think about it. `frameInternal` makes more sense.

Sources/NIOQUIC/GSOCoalescer.swift
glbrntt: No: it doesn't need any metadata, it's a single datagram without GSO.

Sources/NIOQUIC/GSOCoalescer.swift
```diff
@@ -13,55 +13,105 @@
...
+    private mutating func datagram(from frame: consuming Frame) -> AddressedEnvelope<ByteBuffer> {
+        let buffer: ByteBuffer
+
+        if let config = frame.takeOwnershipOfCustomFinalizerBuffer() {
+            frame.finalize(success: true)
+            buffer = ByteBuffer(
+                takingOwnershipOf: config.bufferPointer,
+                allocator: FrameMemory.allocator,
+                readerIndex: config.readerOffset,
+                writerIndex: config.writerOffset
```
glbrntt:
```suggestion
        if let frameInternals = frame.takeOwnershipOfCustomFinalizerBuffer() {
            frame.finalize(success: true)
            buffer = ByteBuffer(
                takingOwnershipOf: frameInternals.bufferPointer,
                allocator: FrameMemory.allocator,
                readerIndex: frameInternals.readerOffset,
                writerIndex: frameInternals.writerOffset
```

rnro (approved):

PR #63, opened 2026-08-01
Fix Conformance to InboundStreamHandler

agnosticdev:
### Motivation

This is more of a warning for when the tag is next bumped up that there was a change that went in recently to SwiftNetwork that will cause an protocol conformance breakage for InboundStreamHandler:

https://github.com/apple/swift-network-evolution/commit/c7adf0f68e243d120a949935b23f9ab82acd0316

This change is a suggested fix for when the time is right.

### Modifications

Conformance for InboundStreamHandler in QUICChannelStreamHandler.

### Result

Builds as expected.

### DIFF
--- a/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
+++ b/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
@@ -1142,7 +1142,7 @@ extension QUICChannelStreamHandler: UpperProtocolHandler {
 }
 
 @available(anyAppleOS 26, *)
-extension QUICChannelStreamHandler: InboundStreamHandler {
+extension QUICChannelStreamHandler {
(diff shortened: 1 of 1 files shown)


glbrntt (approved):
Thanks Matt!

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -1141,6 +1141,16 @@ extension QUICChannelStreamHandler: UpperProtocolHandler {
     }
 }
 
+@available(anyAppleOS 26, *)
+extension QUICChannelStreamHandler: InboundStreamHandler {
```
glbrntt:
```suggestion
extension QUICChannelStreamHandler {
```

PR #64, opened 2026-08-03
Reduced QUICChannelStreamHandler layout by 128 bytes

agnosticdev:
### Motivation

The memory layout of QUICChannelStreamHandler is currently 640 bytes.  This change tweaks some of the properties to knock it down to 512 Bytes.

### Modifications

Instead of storing Duration I stored the UInt16 that the keep alive functionality needed.
Instead of storing the `ByteBufferAllocator` I returned it through connectionChannel in the computed property.
I moved `logPrefix` under a DEBUG flag, since its only used there.  I also added logic to remove the `logPrefix` during `error` and `warning` logs too.

### Result

Less memory to allocated when creating a stream.

### DIFF
--- a/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
+++ b/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
@@ -1167,16 +1167,6 @@ extension QUICChannelStreamHandler: UpperProtocolHandler {
     }
 }
 
-@available(anyAppleOS 26, *)
-extension QUICChannelStreamHandler: InboundStreamHandler {
-    func attachLowerStreamProtocolToExistingFlow(
-        listener: StreamListenerLinkage,
-        flowReference: ProtocolInstanceReference
-    ) throws(NetworkError) {
-        throw NetworkError.posix(ENOTSUP)
-    }
-}
-
 // MARK: - `Channel` and `ChannelCore` conformance
 
 @available(anyAppleOS 26, *)
--- a/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
+++ b/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
@@ -32,7 +32,7 @@ enum StreamClosureState {
     case disconnectOnly  // Calls stop and detach only, could result in disconnect event as well
 }
 
-fileprivate enum LogLevel {
+private enum LogLevel {
     case trace
     case warning
     case error
--- a/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
+++ b/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
@@ -32,6 +32,12 @@ enum StreamClosureState {
     case disconnectOnly  // Calls stop and detach only, could result in disconnect event as well
 }
 
+enum LogLevel: Int {
+    case trace
+    case warning
+    case error
+}
+
 /// `QUICChannelStreamHandler` is the bridge between SwiftNetwork and our code on the application-side;
 /// one `QUICChannelStreamHandler` exists for each QUIC stream.
 @available(anyAppleOS 26, *)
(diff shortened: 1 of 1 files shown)


Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -127,7 +138,9 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
         self._remoteAddress = remoteAddress
         self._localAddress = localAddress
         self.context = parameters.context
-        self.keepAliveInterval = keepAliveInterval
+        if let keepAlive = keepAliveInterval {
+            self._keepAliveInterval = UInt16(truncatingIfNeeded: keepAlive.components.seconds)
```
glbrntt: This doesn't do what you think it does:

```swift
let tooManySeconds = Int(UInt16.max) + 1
let truncated = UInt16(truncatingIfNeeded: tooManySeconds)
print("truncated: \(truncated)")  // truncated: 0
```

You want `UInt16(clamping:)`

```swift
let clamped = UInt16(clamping: tooManySeconds)
print("clamped: \(clamped)")  // clamped: 65535
```

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -32,6 +32,12 @@ enum StreamClosureState {
     case disconnectOnly  // Calls stop and detach only, could result in disconnect event as well
 }
 
+enum LogLevel: Int {
```
glbrntt: Does this need `Int` at all? Also can you make this `fileprivate`?

glbrntt (changes requested):

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
agnosticdev: Addressed in [c306cd2](https://github.com/apple/swift-nio-quic/pull/64/commits/c306cd21aee2fb36455bc4e48fc5eab0f11f40b4)

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
agnosticdev: `truncatingIfNeeded` and going back to 0 will not set the keep alive time, see reference [here](https://github.com/apple/swift-network-evolution/blob/main/Sources/SwiftNetwork/QUIC/QUICConnection.swift#L2853) which I think is fine in this case if you have passed in an invalid time value.  Passing in `UInt16.max` is the actual default keepalive value, see reference here (https://github.com/apple/swift-network-evolution/blob/main/Sources/SwiftNetwork/QUIC/Constants.swift#L45).

glbrntt: As a user I'd find that behavior very surprising! I think clamping to the largest possible keepalive would be less surprising...

agnosticdev: For a better user experience than I can support that case.  Changed in [b070b5f](https://github.com/apple/swift-nio-quic/pull/64/commits/b070b5fae58fbe8d68da9d95d825d4d22ea72d15)

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -32,6 +32,12 @@ enum StreamClosureState {
     case disconnectOnly  // Calls stop and detach only, could result in disconnect event as well
 }
 
+fileprivate enum LogLevel {
```
glbrntt: Formatter wants private... sorry for misleading you 😅

```suggestion
private enum LogLevel {
```

glbrntt (approved):
Thanks Matt – nice patch!

PR #65, opened 2026-08-04
Fire a connection's close error before closing its streams

glbrntt:
Motivation:

This is a behavioural regression from the connection channel rewrite.

Modifications:

- Fire 'errorCaught' in 'fireChannelInactiveNow', before 'closeAllStreams', rather than in 'completeChannelInactive' after it.

Result:

Connection errors are fired before closing streams.

### DIFF
--- a/Sources/NIOQUIC/QUICConnectionChannel.swift
+++ b/Sources/NIOQUIC/QUICConnectionChannel.swift
@@ -762,6 +762,12 @@ extension QUICConnectionChannel {
             ()
         }
 
+        // Surface the close error to before tearing the streams down: a handler may rely on
+        // connection level errors to fail streams.
+        if let error {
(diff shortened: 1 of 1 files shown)


Sources/NIOQUIC/QUICConnectionChannel.swift
```diff
@@ -762,6 +762,12 @@ extension QUICConnectionChannel {
             ()
         }
 
+        // Surface the close error to before tearing the streams down: a handler may rely on
+        // connection level errors to fail streams.
+        if let error {
+            self.pipeline.fireErrorCaught(error)
+        }
+
```
fabianfett, line 770: since this is our own channel implementation, should we have a test for this one? Otherwise LGTM.

fabianfett (approved):
Would love to see the test added.

PR #66, opened 2026-08-04
Always complete a stream channel's close promise

glbrntt:
Motivation:

If a stream is openend in the same read batch that closes the connection, the stream will be created but never complete its close promise. Note that the stream channel won't have been configured.

The connection channel waits for all streams to close before it fires inactive so this results in the connection wedging open.

Modifications:

- Making closing a stream idempotent
- When closing, check whether to the stream was active first and conditionally fire inactive

Result:

Fewer wedges

### DIFF
--- a/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
+++ b/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
@@ -1366,17 +1366,17 @@ extension QUICChannelStreamHandler: Channel, ChannelCore {
         // its close promise must still be completed.
         let wasActive = self._isActive.exchange(false, ordering: .sequentiallyConsistent)
 
-        // Put calls to _close on the next runloop tick because it calls fireChannelInactive
+        if let error {
+            self.pipeline.fireErrorCaught(error)
+        }
+
+        if wasActive {
+            self.pipeline.fireChannelInactive()
+            self.pipeline.fireChannelUnregistered()
+        }
+
+        // Complete the promise on the next loop tick.
         self.eventLoop.assumeIsolated().execute {
-            if let error {
-                self.pipeline.fireErrorCaught(error)
-            }
-
-            if wasActive {
-                self.pipeline.fireChannelInactive()
-                self.pipeline.fireChannelUnregistered()
-            }
-
             self.removeHandlers(pipeline: self.pipeline)
             self._closePromise.succeed(())
             promise?.succeed()
--- a/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
+++ b/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
@@ -118,6 +118,9 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
     /// Defaults to `true`. This value can be updated by calling `setOption`.
     private var autoRead: Bool = true
 
+    /// Whether `_close` has already run.
+    private var isClosed: Bool = false
+
     private var originalMetadata: ProtocolMetadata<QUICProtocol>?
     private var connectedEventHandler: ((QUICStreamID?) -> Void)?
     private var disconnectedEventHandler: ((NetworkError?) -> Void)?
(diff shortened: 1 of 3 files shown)


Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -668,6 +671,11 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
         }
     }
 
+    /// Closes the channel if it isn't yet closed.
+    func closeIfNeeded() {
```
fabianfett, line 675: this is a protocol requirement, right?

fabianfett: Nope. Internal interface to private method. My bad.

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -1345,20 +1352,31 @@ extension QUICChannelStreamHandler: Channel, ChannelCore {
...
+        let wasActive = self._isActive.exchange(false, ordering: .sequentiallyConsistent)
+
         // Put calls to _close on the next runloop tick because it calls fireChannelInactive
         self.eventLoop.assumeIsolated().execute {
             if let error {
                 self.pipeline.fireErrorCaught(error)
             }
 
-            self.pipeline.fireChannelInactive()
-            self.pipeline.fireChannelUnregistered()
```
fabianfett, line 1361: The NIO/Netty rule is promise before channel pipeline, isn't it?

fabianfett (commented):
I _think_, I found an actual small issue. Definitely should get its own test as well.

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
glbrntt: No, I don't think so. Here's http/2 for reference: https://github.com/apple/swift-nio-http2/blob/45bdf670248be5f16ec0340e125dca285536f0fb/Sources/NIOHTTP2/HTTP2StreamChannel.swift#L688-L711

What this does highlight though is that error caught / inactive / unregistered should happen before the `execute`.

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -1345,20 +1352,31 @@ extension QUICChannelStreamHandler: Channel, ChannelCore {
...
+        if wasActive {
             self.pipeline.fireChannelInactive()
             self.pipeline.fireChannelUnregistered()
+        }
+
+        // Complete the promise on the next loop tick.
+        self.eventLoop.assumeIsolated().execute {
             self.removeHandlers(pipeline: self.pipeline)
             self._closePromise.succeed(())
             promise?.succeed()
```
fabianfett, line 1382: wait. but this promise is the close promise has to be succeeded before `fireChannelInactive`.

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -1345,20 +1352,31 @@ extension QUICChannelStreamHandler: Channel, ChannelCore {
...
-            }
+        self.isClosed = true
 
+        // A stream can be closed before it ever became active, e.g. its flow was attached in the
+        // same read batch in which the connection tore down.
+        //
+        // If that happens it must not see 'channelInactive' (it didn't see 'channelActive') but
+        // its close promise must still be completed.
+        let wasActive = self._isActive.exchange(false, ordering: .sequentiallyConsistent)
+
```
fabianfett, line 1368: h2 also has failing pending writes and connects here. do we need to add those as well?

glbrntt: No, there's nothing pending.

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
glbrntt: Right, this was incorrect prior to this change as well. 

I don't want to scope creep this PR and introduce more behavior changes: it was intentionally fixing just one thing. More than happy to fix other behavioral issues up in a different PR.

fabianfett (approved):

PR #69, opened 2026-08-06
Add a frame pool

glbrntt:
Motivation:

Before creating Frames, SwiftNetwork asks us for a batch of Frames to write outbound data into. It then populates them and hands them back to us. At that point we then copy their contents into out pooled ByteBuffer's. We can do the same for Frames: pool them.

Modifiactions:

Add a frame pool which stores two size classes of Frame: small and large. When GSO is used the large frames will be sized at 64KB, the largest size possible for a GSO write. When GSO isn't used the large frames won't be used at all. In both cases small frames will be sized at ~1500 bytes, i.e. larger than the MSS.

The pool has capacity limits on the number of each size class that can be stored to avoid holding memory for too long (e.g. if a burst of traffic results in many frames being created and then stored indefinitely).

Result:

Nothing yet, it isn't used. However, when used it will reduce allocs on the outbound path.

### DIFF
--- a/Sources/NIOQUIC/FramePool.swift
+++ b/Sources/NIOQUIC/FramePool.swift
@@ -1,0 +1,188 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2026 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import BasicContainers
+@_spi(ProtocolProvider) import SwiftNetwork
+
+/// A pool of `Frame`s.
+///
+/// The pool stores two category sizes of frame: small and large. To be pooled, small and large
+/// frames must have exactly the capacity as required by their size class. The pool limits also how
+/// many of each category size can be stored at a given time. This avoids situations where bursty
+/// traffic results in the pool overcomitting memory indefinitely.
+@available(anyAppleOS 26, *)
+final class FramePool {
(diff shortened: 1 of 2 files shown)


Sources/NIOQUIC/FramePool.swift
```diff
@@ -0,0 +1,188 @@
...
+            stored = self.smallStorage.storeBuffer(buffer)
+        case self.largeFrameSize:
+            stored = self.largeStorage.storeBuffer(buffer)
+        default:
+            // Not a pooled size.
+            buffer.deallocate()
+            stored = false
+        }
+
+        return stored
```
fabianfett, line 86: why don't we return from within the switch? the `stored` makes this harder to read, IMO.

glbrntt: I disagree, but this is subjective. 

My rationale for preferring this is that the more control flow paths you add, the higher the cognitive burden is to understanding the code. The higher the cognitive burden the longer it takes to read and understand code.

You could argue that this is simple enough that it doesn't matter - which is a completely reasonable position to take - but the question then becomes where does that line fall? What is simple enough? I choose to not care and write the simplest code where possible and minimise branches.

glbrntt: FWIW: I don't think it makes this harder to read, it just make it more lines of code. If anything it's more explicit: the return value of `storeBuffer` has more meaning when assigned to `stored` than if returned directly.

fabianfett: okay, stupid me this is likely because of exclusive memory access in noncopyable types.

fabianfett (approved):
Looks good to me.

Sources/NIOQUIC/FramePool.swift
glbrntt: No, that shouldn't be relevant here...

glbrntt:
GitHub actions is in a weird state, presumably from the outage. Closing and reopening in the hope it clears the issue.

PR #70, opened 2026-08-06
Propagate batching to SwiftNetwork

glbrntt:
Motivation:

Every flush on a stream channel flushes out all pending frames in the QUIC connection. This is suboptimal because in a read loop there may be N streams that want to write. It's significantly more efficient to write on the N frames and then flush out the connection.

SwiftNetwork just added a signal we can use to hint that outbound writes should be batched. We set this at the start of a read loop and unset it at the end which causes SwiftQUIC to flush out data.

Modifications:

- Bump version
- Wire through outbound batching events
- At the end of the read loop the connection channel now tells QUIC to consume the inbound packets, disables batching (so flushes out writes) and then reconciles state.
- If there are pending streams then batching switched back on again as the init of streams may lead to writes. These are flushed out afterwards.
- Simplification to the processing of inbound streams: it no longer returns a boolean we always flush afterwards.

Result:

Better perf

### DIFF
--- a/Sources/NIOQUIC/QUICConnectionChannel.swift
+++ b/Sources/NIOQUIC/QUICConnectionChannel.swift
@@ -874,28 +874,28 @@ extension QUICConnectionChannel {
     fileprivate func _parentChannelReadComplete() {
         self.eventLoop.assertInEventLoop()
 
-        defer {
-            switch self._connection {
-            case .live(let connection):
-                connection.exitReadLoop()
-            case .test:
-                ()
-            }
-        }
-
         // Avoid entering 'drainOutput'; wait for all events to be delivered and then
         // deal with them in 'drainAndReconcileLifecycle' below.
         self.withoutEnteringDrainOutput {
             self._connection.receivePacketsComplete()
         }
 
-        self.drainAndReconcileLifecycle()
-        let initializedAnyStreams = self.processPendingInboundStreams()
-
-        // Stream initializers may produce output; reconcile any pending events again.
-        if initializedAnyStreams {
-            self.drainAndReconcileLifecycle()
+        // Disable outbound batching so that SwiftQUIC emits frames into the connection. They'll
+        // be picked up when draining starts a few lines below. Re-enable again (i.e. disable
+        // temporarily) if there are pending streams as they may also produce output during their
+        // init.
+        self._connection.withLiveOnly { connection in
+            let hasPendingStreams = !self._pendingStreams.isEmpty
+            connection.flushOutboundBatch(resumeBatching: hasPendingStreams)
         }
+
+        self.drainAndReconcileLifecycle()
+        self.processPendingInboundStreams()
+
+        // Done producing data: exit read loop (and implicitly disables outbound batching).
+        self._connection.withLiveOnly { $0.exitReadLoop() }
+        // Stream initializers may produce output; reconcile any pending events again.
+        self.drainAndReconcileLifecycle()
     }
 
     fileprivate func _parentChannelWritabilityChanged(to isWritable: Bool) {
@@ -1001,19 +1001,15 @@ extension QUICConnectionChannel {
     }
 
     /// Initialize inbound streams pushed during the current read batch, skipping any not yet
-    /// connected. Returns `true` if any stream was initialized.
-    private func processPendingInboundStreams() -> Bool {
+    /// connected.
+    private func processPendingInboundStreams() {
         self.eventLoop.assertInEventLoop()
-        var anyStreamInitialized = false
 
         while let (streamID, stream) = self._pendingStreams.popFirst() {
             if stream.streamStateMachine.isConnected {
                 self.runInboundStreamInitializer(streamID: streamID, stream: stream)
-                anyStreamInitialized = true
             }
         }
-
-        return anyStreamInitialized
     }
 
     private func runInboundStreamInitializer(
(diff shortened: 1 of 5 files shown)


Sources/NIOQUIC/SwiftNetwork/SwiftNetworkQUICConnection.swift
```diff
@@ -90,10 +90,32 @@ final class SwiftNetworkQUICConnection {
...
+        self.setOutboundBatching(false)
         self.inReadLoop = false
     }
 
+    internal func setOutboundBatching(_ enabled: Bool) {
+        if self.batchingEnabled == enabled { return }
+        guard let newFlowHandler = self.connectionNewFlowHandler else { return }
+
+        self.batchingEnabled = enabled
+        newFlowHandler.outboundBatching(enabled)
```
fabianfett, line 112: this call will likely trigger another method call on this object I assume? can we add a quick comment/hint what will be called as a sideeffect of this call?

Sources/NIOQUIC/SwiftNetwork/SwiftNetworkQUICConnection.swift
```diff
@@ -874,7 +897,12 @@ final class SwiftNetworkQUICConnection {
     @discardableResult
     @inlinable
     func receivePacket(_ packet: NIOCore.ByteBuffer) -> Int {
-        self.inReadLoop = true
+        if !self.inReadLoop {
+            self.inReadLoop = true
+            self.batchingEnabled = true
+            self.connectionNewFlowHandler?.outboundBatching(true)
```
fabianfett: why isn't this calling `self.setOutboundBatching(true)`?

Sources/NIOQUIC/QUICConnectionChannel.swift
```diff
@@ -870,28 +870,28 @@ extension QUICConnectionChannel {
...
+        // temporarily) if there are pending streams as they may also produce output during their
+        // init.
+        self._connection.withLiveOnly { connection in
+            let hasPendingStreams = !self._pendingStreams.isEmpty
+            connection.flushOutboundBatch(resumeBatching: hasPendingStreams)
+        }
+
         self.drainAndReconcileLifecycle()
-        let initializedAnyStreams = self.processPendingInboundStreams()
+        self.processPendingInboundStreams()
```
fabianfett, line 889: based on the comment above shouldn't we flush another time after all the pending inbound streams have been resolved? if not, can we add a comment, how we reconcile the in batch state with not being in a read loop?

fabianfett (approved):

Sources/NIOQUIC/SwiftNetwork/SwiftNetworkQUICConnection.swift
glbrntt: Oversight! I refactored this last night to add `setOutboundBatching` but clearly missed this call site. Will fix up.

Sources/NIOQUIC/QUICConnectionChannel.swift
glbrntt: We do: `exitReadLoop` disables batching again and the `drainAndReconcileLifecycle` below that will pull outbound datagrams from the connection and write them into the UDP channel.

I think this is another victim of my last minute renaming/refactor: the comment below says it disables outbound batching rather than "flush outbound batch" (which is the same thing). I'll update.

Sources/NIOQUIC/SwiftNetwork/SwiftNetworkQUICConnection.swift
```diff
@@ -106,6 +106,9 @@ final class SwiftNetworkQUICConnection {
         guard let newFlowHandler = self.connectionNewFlowHandler else { return }
 
         self.batchingEnabled = enabled
+        // Enabling batching stops the underlying connection from producing (most) datagrams until
+        // batching is disabled again, at which point any queued outbound writes will be emitted.
+        // This allows outbound wrties to be better packed into datagrams.
```
fabianfett:
```suggestion
        // This allows outbound writes to be better packed into datagrams.
```

fabianfett: Can we add where the `SwiftNetwork.QuicConnection` will call into `NIOQUIC` again? Basically if I call this where can I set a breakpoint?

Sources/NIOQUIC/SwiftNetwork/SwiftNetworkQUICConnection.swift
```diff
@@ -90,10 +90,35 @@ final class SwiftNetworkQUICConnection {
...
 
+    internal func setOutboundBatching(_ enabled: Bool) {
+        if self.batchingEnabled == enabled { return }
+        guard let newFlowHandler = self.connectionNewFlowHandler else { return }
+
+        self.batchingEnabled = enabled
+        // Enabling batching stops the underlying connection from producing (most) datagrams until
+        // batching is disabled again, at which point any queued outbound writes will be emitted.
+        // This allows outbound writes to be better packed into datagrams.
+        newFlowHandler.outboundBatching(enabled)
```
glbrntt, line 112:
```suggestion
        // batching is disabled again, at which point any queued outbound writes will be emitted.
        // This allows outbound writes to be better packed into datagrams. When the batch is flushed
        // the connection will receive frames in `outputHandlerFinalizeOutputFrames`.
        newFlowHandler.outboundBatching(enabled)
```

PR #71, opened 2026-08-07
Wire up frame pool

glbrntt:
Motivation:

The output handler creates frames to pass into SwiftNetwork, and the GSOCoalescer consumes those frames. Instead of throwing them on the floor we can pool them in a FramePool.

Modifications:

- Wire up the frame pool

Result:

Fewer allocations

### DIFF
--- a/Sources/NIOQUIC/GSOCoalescer.swift
+++ b/Sources/NIOQUIC/GSOCoalescer.swift
@@ -35,8 +35,12 @@ struct GSOCoalescer: ~Copyable {
     /// A pool of byte buffers.
     private var pool: BufferPool
 
+    /// A pool of `Frame`s.
+    private let framePool: FramePool
+
     init(
         remoteAddress: SocketAddress,
+        framePool: FramePool,
         maxSegments: Int = 64,
         maxCoalescedSize: Int = 65535,
         bufferPoolCapacity: Int = 8
@@ -49,6 +53,7 @@ struct GSOCoalescer: ~Copyable {
         self.frames.reserveCapacity(16)
 
         self.pool = BufferPool(capacity: bufferPoolCapacity, allocator: ByteBufferAllocator())
+        self.framePool = framePool
     }
 
     var isEmpty: Bool {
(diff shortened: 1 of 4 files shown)


Sources/NIOQUIC/GSOCoalescer.swift
```diff
@@ -144,10 +139,7 @@ struct GSOCoalescer: ~Copyable {
             let (buffer, ()) = self.pool.withBuffer(minimumCapacity: totalSize) { buffer in
                 while runLength > 0 {
                     runLength &-= 1
-
-                    var frame = frames.popFirst()!
-                    frame.span?.withUnsafeBufferPointer { _ = buffer.writeBytes($0) }
-                    frame.finalize(success: true)
+                    Self.write(self.frames.popFirst()!, to: &buffer, returningFrameTo: self.framePool)
```
fabianfett: Since we are already using unsafe math, do we consider faster unwrapping of the optional?
```suggestion
                    Self.write(self.frames.popFirst().unsafelyUnwrapped, to: &buffer, returningFrameTo: self.framePool)
```

glbrntt: Technically the maths here isn't "unsafe"; it's "overflow" arithmetic but that's neither here nor there.

Hmm. I'm not so sure as we generally don't have a strong position on `unsafelyUnwrapped`, I'm not sure I've ever seen it in any of our codebases. I also think it's harder to reason about than the arithmetic here since it relies on the `runLength` computation being correct (i.e. `frames` has enough elements in) so I'd be a bit hesitant about using an unsafe spelling here. 

The better spelling here is `removeFirst()` I think.

glbrntt: (Side note: the `Self.write` func here is an annoying artifact of the `~Copyable` frame.)

fabianfett: We use `unsafelyUnwrapped` in PostgresNIO when we have done bounds checks before, to eliminate forks in the fast path.

fabianfett (approved):

PR #72, opened 2026-08-10
Various fixes

glbrntt:
### Motivation

The connection channel refactor surfaced a few regression in HTTP/3. These were mostly hidden because of the stream channel not behaving as a regular NIO channel should. One of these behaviors was corrected in #66 but surfaced even more issues. This change fixes a number of small issues.

### Modifications

- Complete a user supplied close promise before firing channel inactive
- Don't swallow pending end-of-stream 
- Only record the first pending close reason 
- Avoid surfacing input closed more than once

### Result

HTTP/3 tests pass

### DIFF
--- a/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
+++ b/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
@@ -718,7 +718,7 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
     /// advancing its flow control window, defeating backpressure and letting unbounded bytes accumulate here.
     private func deliverInboundDataDownstreamIfDemand() {
         // Only continue if there is a pending read.
-        guard self.pendingRead && !self.isClosed else {
+        guard self.pendingRead else {
             return
         }
 
--- a/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
+++ b/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
@@ -692,7 +692,8 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
             break
         }
 
-        self._close(error: error, promise: nil)
+        let quicError = error.flatMap { QUICConnectionError(networkError: $0) }
+        self._close(error: quicError ?? error, promise: nil)
     }
 
     @inline(__always)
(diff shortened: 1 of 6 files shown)


Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
```diff
@@ -717,7 +718,7 @@ final class QUICChannelStreamHandler: ProtocolInstanceContainer, InboundStreamHa
     /// advancing its flow control window, defeating backpressure and letting unbounded bytes accumulate here.
     private func deliverInboundDataDownstreamIfDemand() {
         // Only continue if there is a pending read.
```
fabianfett, line 720:
```suggestion
        // Only continue if there is a pending read and we are not closed.
```

fabianfett: what is the code path to call this, even when already closed?

Sources/NIOQUIC/SwiftNetwork/QUICStreamStateMachine.swift
```diff
@@ -671,10 +667,12 @@ struct QUICStreamStateMachine: ~Copyable {
         /// Shut down both directions — caller should stop and remove handler.
         case shutdownBoth
         /// Read side closed. If `streamFullyClosed`, stop and remove handler;
-        /// otherwise send STOP_SENDING via `abortInbound`.
-        case closeRead(streamFullyClosed: Bool)
-        /// Read side was already closed — no action needed.
-        case readAlreadyClosed
+        /// otherwise send STOP\_SENDING via `abortInbound`. When `deliverEndOfStream`
```
fabianfett, line 670: why the additional backslash?
```suggestion
        /// otherwise send STOP_SENDING via `abortInbound`. When `deliverEndOfStream`
```

fabianfett (approved):

Sources/NIOQUIC/SwiftNetwork/QUICStreamStateMachine.swift
glbrntt: The `_` is treated as a delimiter for italics in DocC (well, markdown strictly speaking) so the backslash escapes that to avoid incorrectly rendering the next chunk in italics.

Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift
glbrntt: Good question. It shouldn't actually be necessary here: we'll only hit this path in when SwiftNetwork calls us back with data which shouldn't happen after the stream has closed.

PR #74, opened 2026-08-18
Add a custom key-value store for QUICStreamID keys

glbrntt:
Motivation:

The cost of hashing stream IDs is non-trivial on the hot path. We can use our knowledge of how stream IDs are structured and used to build a faster, custom data structure.

Modifications:

Add QUICStreamIDDictionary: a key-value store which only allows QUICStreamID to be useds as a key type.

This is built on top of QUICStreamIDCache which is a dynamically resizing array of optional stream ID and value pairs. The slot index is derived from the numeric bits of the stream ID modulo the capacity of the cache. The dictionary holds four of these caches (one per stream type) and an overflow dictionary. When a collison occurs in the cache (i.e. because a stream is long lived) then the older value is evicted and moved to the overflow dictionary.

This works quite nicely because the cache tends towards the number of open streams of that type and they effectively round-robin. Checking the cache for a value is cheap: a mask to lookup which cache, and then a bit shift and mask to lookup the slot within the cache.

Result:

Better perf.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ DerivedData/
 .swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
 .netrc
 Package.resolved
+Benchmarks/.build
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,4 @@ DerivedData/
 .swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
 .netrc
 Package.resolved
-Benchmarks/.build
+.build
--- a/Benchmarks/Benchmarks/QUICStreamIDDictionaryBenchmarks/QUICStreamIDCache.swift
+++ b/Benchmarks/Benchmarks/QUICStreamIDDictionaryBenchmarks/QUICStreamIDCache.swift
@@ -1,0 +1,1 @@
+../../../Sources/NIOQUIC/QUICStreamIDCache.swift
--- a/Benchmarks/Benchmarks/QUICStreamIDDictionaryBenchmarks/QUICStreamIDDictionary.swift
+++ b/Benchmarks/Benchmarks/QUICStreamIDDictionaryBenchmarks/QUICStreamIDDictionary.swift
@@ -1,0 +1,1 @@
+../../../Sources/NIOQUIC/QUICStreamIDDictionary.swift
--- a/Sources/NIOQUIC/QUICStreamIDCache.swift
+++ b/Sources/NIOQUIC/QUICStreamIDCache.swift
@@ -1,0 +1,219 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2026 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOQUICHelpers
+
+/// A cache of values keyed by `QUICStreamID`. See ``QUICStreamIDDictionary``.
+///
+/// Slots are indexed by the numeric part of a stream ID (i.e. `rawValue >> 2`) modulo the capacity
+/// of the cache, which is always a power of two so that the modulo can be done by masking.
+struct QUICStreamIDCache<Value> {
+    fileprivate struct Entry {
+        var id: QUICStreamID
+        var value: Value
+    }
+
+    /// The underlying slots in the cache, `nil` means the slot is free.
+    private var slots: [Entry?]
+
+    /// A mask applied to the numeric part of a stream ID (i.e. top 62 bits) to get its slot index.
+    /// Stored rather than recomputed to avoid the `Int` to `UInt64` conversion on every lookup.
+    private var mask: UInt64
+
+    /// The value of `count` at which the cache doubles in size.
+    private var nextGrowthCount: Int
+
+    /// The utilisation threshold above which the cache will double in size.
+    private let threshold: Double
+
+    /// The number of elements currently stored in the cache.
+    private(set) var count: Int
+
+    /// The number of elements that can be stored in the cache.
+    var capacity: Int { self.slots.count }
+
+    /// Whether the cache is empty.
+    var isEmpty: Bool { self.count == 0 }
+
+    init(capacity: Int, threshold: Double) {
+        let capacity = capacity.nextPowerOfTwo
+        self.slots = Array(repeating: nil, count: capacity)
+        self.mask = UInt64(capacity - 1)
+        self.threshold = threshold
+        self.nextGrowthCount = capacity.scaled(by: threshold)
+        self.count = 0
+    }
+
+    /// Index of the slot for the given stream ID.
+    func slotIndex(of id: QUICStreamID) -> Int {
+        // Drop the type bits and then mask. The mask can be used instead of '%' as the capacity is
+        // guaranteed to be a power of two (and the mask is just `capacity - 1`).
+        Int((id.rawValue >> 2) & self.mask)
+    }
+
+    /// Returns the value for the given stream ID, if it exists in the cache.
+    subscript(id: QUICStreamID) -> Value? {
+        if let entry = self.slots[self.slotIndex(of: id)], entry.id == id {
+            return entry.value
+        } else {
+            return nil
+        }
+    }
+
+    /// Returns whether the cache contains the given stream ID.
+    func contains(_ id: QUICStreamID) -> Bool {
+        self.slots[self.slotIndex(of: id)]?.id == id
+    }
+
+    enum UpdateResult {
+        /// The value was inserted into an empty slot.
+        case inserted
+        /// The value replaced a value for the same stream ID.
+        case replaced(Value)
+        /// The value was inserted but evicted a value for a different stream ID.
+        case evicted(id: QUICStreamID, value: Value)
+    }
+
+    /// Updates the value stored for the given stream ID.
+    ///
+    /// - Parameters:
+    ///   - value: The value to store.
+    ///   - id: The ID of the stream.
+    /// - Returns: Whether the value was inserted, replaced an existed value, or evicted a value
+    ///   for another stream.
+    @discardableResult
+    mutating func updateValue(_ value: Value, forID id: QUICStreamID) -> UpdateResult {
+        let index = self.slotIndex(of: id)
+
+        var entry: Entry? = Entry(id: id, value: value)
+        swap(&self.slots[index], &entry)
+
+        switch entry {
+        case .none:
+            self.count &+= 1
+            self.doubleCapacityIfNeeded()
+            return .inserted
+
+        case .some(let previous):
+            if previous.id == id {
+                return .replaced(previous.value)
+            } else {
+                return .evicted(id: previous.id, value: previous.value)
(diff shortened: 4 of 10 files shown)


Sources/NIOQUIC/QUICStreamIDCache.swift
```diff
@@ -0,0 +1,219 @@
...
+
+    /// The number of elements that can be stored in the cache.
+    var capacity: Int { self.slots.count }
+
+    /// Whether the cache is empty.
+    var isEmpty: Bool { self.count == 0 }
+
+    init(capacity: Int, threshold: Double) {
+        let capacity = capacity.nextPowerOfTwo
+        self.slots = Array(repeating: nil, count: capacity)
```
agnosticdev, line 52: Just a thought. Could you start with a know size here and make this an InlineArray and then you could put these `slots` on the stack?  Or does it make more sense to dynamically grow these structures?

Sources/NIOQUIC/QUICStreamIDDictionary.swift
```diff
@@ -0,0 +1,234 @@
...
+    /// An overflow dictionary for when a value is evicted from a cache.
+    private var overflow: [QUICStreamID: Value]
+
+    /// Caches keyed by the "type bits" of a QUIC stream ID (i.e. `rawValue & 0b11`).
+    private var caches: InlineArray<4, QUICStreamIDCache<Value>>
+
+    /// Returns the cache index to use for a given stream ID.
+    private func cacheIndex(of id: QUICStreamID) -> Int {
+        // Bottom two bits are the type bits.
+        Int(id.rawValue & 0b11)
```
agnosticdev, line 68: Nice!

Sources/NIOQUIC/QUICStreamIDDictionary.swift
```diff
@@ -0,0 +1,234 @@
...
+/// dictionary). Each cache is separately sized and doubles in capacity when its load factor (ratio
+/// of occupied slots to capacity) exceeds a threshold. This is naturally bounded by the connections
+/// limit on concurrent streams.
+@available(anyAppleOS 26, *)
+struct QUICStreamIDDictionary<Value> {
+    /// An overflow dictionary for when a value is evicted from a cache.
+    private var overflow: [QUICStreamID: Value]
+
+    /// Caches keyed by the "type bits" of a QUIC stream ID (i.e. `rawValue & 0b11`).
+    private var caches: InlineArray<4, QUICStreamIDCache<Value>>
```
agnosticdev, line 37: (Just for the sake that I do in-fact understand; this is one cache for each stream type correct?)

agnosticdev (approved):
This is really cool, nice work! One just lingering though I had was if there was any overhead in CPU cost you see from maintaining this logic?

Sources/NIOQUIC/QUICStreamIDCache.swift
```diff
@@ -0,0 +1,219 @@
...
+    /// The number of elements currently stored in the cache.
+    private(set) var count: Int
+
+    /// The number of elements that can be stored in the cache.
+    var capacity: Int { self.slots.count }
+
+    /// Whether the cache is empty.
+    var isEmpty: Bool { self.count == 0 }
+
+    init(capacity: Int, threshold: Double) {
```
fabianfett, line 49: should the threshold be between 0 and 1? should we precondition enforce that?

Sources/NIOQUIC/QUICStreamIDCache.swift
```diff
@@ -0,0 +1,219 @@
...
+        let capacity = capacity.nextPowerOfTwo
+        self.slots = Array(repeating: nil, count: capacity)
+        self.mask = UInt64(capacity - 1)
+        self.threshold = threshold
+        self.nextGrowthCount = capacity.scaled(by: threshold)
+        self.count = 0
+    }
+
+    /// Index of the slot for the given stream ID.
+    func slotIndex(of id: QUICStreamID) -> Int {
```
fabianfett, line 60: OOI can there be a streamID `0`? Basically `0xx` where xx is the stream type marker?

Sources/NIOQUIC/QUICStreamIDDictionary.swift
fabianfett: yes

Sources/NIOQUIC/QUICStreamIDDictionary.swift
```diff
@@ -0,0 +1,234 @@
...
+        Int(id.rawValue & 0b11)
+    }
+
+    /// Returns the number of elements in the dictionary.
+    var count: Int {
+        var total = self.overflow.count
+        total &+= self.caches[0].count
+        total &+= self.caches[1].count
+        total &+= self.caches[2].count
+        total &+= self.caches[3].count
```
fabianfett: should we add a comment, why overflow math is fine here?

Sources/NIOQUIC/QUICStreamIDDictionary.swift
```diff
@@ -0,0 +1,234 @@
...
+    mutating func updateValue(_ value: Value, forID id: QUICStreamID) -> Value? {
+        let previous: Value?
+
+        if self.overflow.isEmpty {
+            // Fast-path: no-overflow values.
+            previous = self.insert(value, forID: id)
+        } else if let index = self.overflow.index(forKey: id) {
+            // Value was in the overflow storage.
+            previous = self.overflow.values[index]
+            self.overflow.values[index] = value
```
fabianfett: nit: its sad that we always need to hash before updating, if there is just a single element in overflow...

Tests/NIOQUICTests/QUICStreamIDDictionaryTests.swift
```diff
@@ -0,0 +1,284 @@
...
+
+import NIOQUICHelpers
+import Testing
+
+@testable import NIOQUIC
+
+struct QUICStreamIDDictionaryTests {
+    @Test
+    @available(anyAppleOS 26, *)
+    func empty() {
```
fabianfett:
```suggestion
    @available(anyAppleOS 26, *)
    @Test func empty() {
```
can we get the `@Test` on the same line as the test name, as this allows VSCode and Xcode single test selection, even when scrolled.

.gitignore
```diff
@@ -7,3 +7,4 @@ DerivedData/
 .swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
 .netrc
 Package.resolved
+Benchmarks/.build
```
fabianfett: should wo go for `.build` - so that it is always ignored, even when fuzzing around with other temp folder?

Benchmarks/Benchmarks/QUICStreamIDDictionaryBenchmarks/QUICStreamIDCache.swift
```diff
@@ -0,0 +1 @@
+../../../Sources/NIOQUIC/QUICStreamIDCache.swift
```
fabianfett, line 1: can we depend on the parent package instead please?

Benchmarks/Benchmarks/QUICStreamIDDictionaryBenchmarks/QUICStreamIDDictionary.swift
```diff
@@ -0,0 +1 @@
+../../../Sources/NIOQUIC/QUICStreamIDDictionary.swift
```
fabianfett, line 1: can we depend on the parent package instead please?

fabianfett (approved):
Generally this looks awesome. A couple of comments.

Benchmarks/Benchmarks/QUICStreamIDDictionaryBenchmarks/QUICStreamIDCache.swift
glbrntt: The types we're testing here are internal. This felt like the least bad option but I'm open to suggestions.

Sources/NIOQUIC/QUICStreamIDCache.swift
glbrntt: Ah yeah, we totally should. We could also get rid of threshold: I benchmarked a bunch of different values and 0.6 was a good tradeoff.

Sources/NIOQUIC/QUICStreamIDCache.swift
glbrntt: I think it's best they dynamically grow: they'll end up sized roughly to the number of concurrent streams of that type.

Sources/NIOQUIC/QUICStreamIDCache.swift
glbrntt: yes

Sources/NIOQUIC/QUICStreamIDDictionary.swift
glbrntt: Yeah... it's a little unfortunate. I might extend this in the future to add an overflow array or something as an intermediate.

Tests/NIOQUICTests/QUICStreamIDDictionaryTests.swift
glbrntt: That's really dumb: Xcode and VSCode should be better and not have to require this...

Sources/NIOQUIC/QUICStreamIDDictionary.swift
```diff
@@ -111,13 +141,13 @@ struct QUICStreamIDDictionary<Value> {
     mutating func updateValue(_ value: Value, forID id: QUICStreamID) -> Value? {
         let previous: Value?
 
-        if self.overflow.isEmpty {
+        if self.overflowCount == 0 {
             // Fast-path: no-overflow values.
             previous = self.insert(value, forID: id)
-        } else if let index = self.overflow.index(forKey: id) {
+        } else if let position = self.overflowIndex(of: id) {
```
fabianfett, line 147: in remove you first check the cache and only then go to overflow, you do this by validating that the streamID matches when accessing the cache. Can't we do the same here?

Sources/NIOQUIC/QUICStreamIDDictionary.swift
```diff
@@ -131,41 +161,142 @@ struct QUICStreamIDDictionary<Value> {
...
+        case .array(let index):
+            self.overflowCount &-= 1
+            // Order has no meaning: swap with the final element to make removal O(1).
+            let lastIndex = self.overflowArray.index(before: self.overflowArray.endIndex)
+            self.overflowArray.swapAt(index, lastIndex)
+            return self.overflowArray.removeLast().value
+
+        case .dictionary(let index):
+            self.overflowCount &-= 1
+            return self.overflowDictionary.remove(at: index).value
```
fabianfett, line 295: okay, so we would only go back to the array approach, if all elements are removed from the dict, correct?

fabianfett (approved):
Love it all! Two small nits to check.

Sources/NIOQUIC/QUICStreamIDDictionary.swift
glbrntt: Correct. I tried switching back when the dictionary drops down to 8 but that was slower – you can end up just bouncing between the two.

Sources/NIOQUIC/QUICStreamIDDictionary.swift
glbrntt: I did benchmark this and it's not a clear win. It's a tiny bit quicker in a few scenarios but slower in more. I think because the containment check in the cache has to load the value. It might be interesting as a future optimization to move the cache to having two parallel arrays, one for stream IDs and the other for values.

PR #75, opened 2026-08-21
Avoid copies and scans in the stream ID dictionary

glbrntt:
Motivation:

The entries in stream ID cache are pairs of ID and value stored as an optional in an array. Reading from the cache means checking for presence first -- and copying the value -- even if just to check a stream ID.

Modifications:

- Replace `Entry?` with a `Slot` which holds an optional `Value` and an always present `UInt64` for the raw stream ID. QUIC stream IDs are only 62-bits in size, so its safe to use `UInt64.max` as a sentinel value.
- Now that contains is cheaper a fast path in `QUICStreamIDDictionary.updateValue(_:forID:)` is available

Result:

- Fewer copies, fewer scans.
- ~5-10% fewer instructions


fabianfett (approved):
LOVE IT

PR #76, opened 2026-08-21
Avoid channel lifecycle changes during oob writes

josephnoir:
### Motivation

Out-of-band writes should not trigger channel lifecycle changes to avoid reentrancy. As an example a channel that turns active might open streams, which calls back into SwiftNetwork.

### Modifications

Restrict `drainOutbound` to draining the output.

### Results

Avoid potential crashes.

### DIFF
--- a/Sources/NIOQUIC/QUICConnectionChannel.swift
+++ b/Sources/NIOQUIC/QUICConnectionChannel.swift
@@ -431,14 +431,14 @@ extension QUICConnectionChannel.ConnectionView {
     }
 
     /// Notifies the connection that it should call back into the connection and drain its output
-    /// buffer.
-    func drainOutbound() {
+    /// buffer. This is used for tests.
+    func drainOutboundAndReconcileLifecycle() {
         self._channel.drainAndReconcileLifecycle()
     }
 
     /// Notifies the connection that it should call back into the connection and drain its output
     /// buffer. This is used for out-of-band writes.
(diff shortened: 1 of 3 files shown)


Sources/NIOQUIC/QUICConnectionChannel.swift
```diff
@@ -430,10 +430,16 @@ extension QUICConnectionChannel.ConnectionView {
...
+    /// buffer. This is used for tests.
+    func drainOutboundAndReconcileLifecycle() {
+        self._channel.drainAndReconcileLifecycle()
+    }
+
     /// Notifies the connection that it should call back into the connection and drain its output
     /// buffer. This is used for out-of-band writes.
     func drainOutbound() {
-        self._channel.drainAndReconcileLifecycle()
+        self._channel.drainOutput()
```
glbrntt: When do the lifecycle changes get applied? Maybe we need to do drain and reconcile on the next loop tick for out-of-band writes?

josephnoir: With this change, they would be applied on the next in-band read or write. Would you move both to the next tick or just the reconciliation?

glbrntt: I think it'd happen _after_ the next read or write, right? If so I think it'd be better to apply them both on the next loop tick. WDYT?

glbrntt (approved):

PR #77, opened 2026-08-24
Only propagate read complete to connections which read

glbrntt:
Motivation:

At the moment the QUICHandler notifies all connections when the UDP channel receives a read complete event. This is wasteful for QUIC channels which handle many connections as only a handlful are likely to read anything.

It's also surprising (but likely inoccuous) behavior for connection channel's to receive: a read complete without a read is a little odd and wastes a pipeline traversal.

Modifications:

- parentChannelRead(_:) returns whether the read was the start of a read loop for a given connection: this allows the QUICHandler to cheaply track which connections are due a read-complete.
- Only notify channels which read at least one packet.

Result:

- Less work.
- ~10% uplift in perf for tests with 1000 concurrent connections

### DIFF
--- a/Sources/NIOQUIC/ConnectionRegistry.swift
+++ b/Sources/NIOQUIC/ConnectionRegistry.swift
@@ -190,12 +190,6 @@ extension ConnectionRegistry where Value == QUICConnectionChannel.TransportView 
         }
     }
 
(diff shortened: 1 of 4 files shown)


josephnoir (approved):
Thanks!

PR #79, opened 2026-08-26
Fire channelReadComplete after reading datagrams

josephnoir:
### Motivation

After reading the available data in a tick channel handleres might expect a read complete call to notify that all data has been read. QUIC datagrams did not respect this and only ever read the datagrams.

### Modifications

If at least one datagrams was read, fire channelReadComplete on the connection channel.

### Results

Better behavior.

### DIFF
--- a/Sources/NIOQUIC/QUICConnectionChannel.swift
+++ b/Sources/NIOQUIC/QUICConnectionChannel.swift
@@ -91,7 +91,10 @@ final class QUICConnectionChannel: @unchecked Sendable {
     /// at the end of the parent channel's read loop (in `_parentChannelReadComplete`).
     private var _pendingStreams: Deque<(QUICStreamID, QUICChannelStreamHandler)>
 
-    /// After a datagram was read during a read loop the channel should fire read complete.
+    /// This flag tracks if QUIC datagrams were forwarded inbound on the connection channel during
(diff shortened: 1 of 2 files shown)


Sources/NIOQUIC/QUICConnectionChannel.swift
```diff
@@ -908,6 +913,12 @@ extension QUICConnectionChannel {
         self.drainAndReconcileLifecycle()
         self.processPendingInboundStreams()
 
+        // If we read a datagram, we should fire channelReadComplete on the connection channel.
```
glbrntt: This is kinda redundant: we're in read complete and the connection channel. What you're really saying is only fire read complete if the channel wrote a datagram.

Sources/NIOQUIC/QUICConnectionChannel.swift
```diff
@@ -91,6 +91,9 @@ final class QUICConnectionChannel: @unchecked Sendable {
     /// at the end of the parent channel's read loop (in `_parentChannelReadComplete`).
     private var _pendingStreams: Deque<(QUICStreamID, QUICChannelStreamHandler)>
 
+    /// After a datagram was read during a read loop the channel should fire read complete.
```
glbrntt: This doesn't really describe what the this bool is for.

glbrntt (approved):
Love it!

aryan-25 (approved):
Thank you!

PR #81, opened 2026-08-27
Add stateless reset token generator

josephnoir:
### Motivation

QUIC stateless reset requires token generation based off a key and a QUIC connection ID. Given the same key and the same connection ID, the generated tokens should be equivalent. Changing either should produce different tokens.

### Modifications

Add a token generator and respective tests.

### Results

Nothing is wired up, this sets up a follow-up PR.

### DIFF
--- a/Sources/NIOQUIC/SwiftNetwork/QUICStatelessResetTokenGenerator.swift
+++ b/Sources/NIOQUIC/SwiftNetwork/QUICStatelessResetTokenGenerator.swift
@@ -51,7 +51,7 @@ struct QUICStatelessResetTokenGenerator: Sendable {
     /// The stateless reset token for `connectionID`.
     func token(for connectionID: QUICConnectionID) -> QUICStatelessResetToken {
         let code = connectionID.withUnsafeBufferPointer {
-            return HMAC<SHA256>.authenticationCode(for: $0, using: self.key)
+            HMAC<SHA256>.authenticationCode(for: $0, using: self.key)
         }
         let token = InlineArray<16, UInt8>(initializingWith: { outputSpan in
             for elem in code.prefix(Self.tokenLength) {
--- a/Sources/NIOQUIC/SwiftNetwork/QUICStatelessResetTokenGenerator.swift
+++ b/Sources/NIOQUIC/SwiftNetwork/QUICStatelessResetTokenGenerator.swift
@@ -50,17 +50,15 @@ struct QUICStatelessResetTokenGenerator: Sendable {
 
     /// The stateless reset token for `connectionID`.
     func token(for connectionID: QUICConnectionID) -> QUICStatelessResetToken {
-        // The initializer only rejects tokens which aren't `tokenLength` bytes long.
-        // Since the RFC defines the token length, we are safe here.
-        QUICStatelessResetToken(self.tokenBytes(for: connectionID))!
-    }
-
-    /// Generate the stateless reset token for `connectionID`.
-    func tokenBytes(for connectionID: QUICConnectionID) -> [UInt8] {
         let code = connectionID.withUnsafeBufferPointer {
             return HMAC<SHA256>.authenticationCode(for: $0, using: self.key)
         }
-        return Array(code.prefix(Self.tokenLength))
+        let token = InlineArray<16, UInt8>(initializingWith: { outputSpan in
+            for elem in code.prefix(Self.tokenLength) {
+                outputSpan.append(elem)
+            }
+        })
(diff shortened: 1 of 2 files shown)


Sources/NIOQUIC/SwiftNetwork/QUICStatelessResetTokenGenerator.swift
```diff
@@ -0,0 +1,92 @@
...
+    func token(for connectionID: QUICConnectionID) -> QUICStatelessResetToken {
+        // The initializer only rejects tokens which aren't `tokenLength` bytes long.
+        // Since the RFC defines the token length, we are safe here.
+        QUICStatelessResetToken(self.tokenBytes(for: connectionID))!
+    }
+
+    /// Generate the stateless reset token for `connectionID`.
+    func tokenBytes(for connectionID: QUICConnectionID) -> [UInt8] {
+        let connectionIDBytes = connectionID.withUnsafeBufferPointer { Array($0) }
+        let code = HMAC<SHA256>.authenticationCode(for: connectionIDBytes, using: self.key)
```
glbrntt: does `authenticationCode(for:using:)` take `[UInt8]`? Or is it `some Collection<UInt8>`? Just wondering if we can easily avoid the extra `Array` alloc on L60 and do something like:

```swift
        let code = connectionID.withUnsafeBufferPointer {
            HMAC<SHA256>.authenticationCode(for: $0, using: self.key)
        }
```

Sources/NIOQUIC/SwiftNetwork/QUICStatelessResetTokenGenerator.swift
```diff
@@ -0,0 +1,92 @@
...
+        triggeringPacketLength: Int,
+        allocator: ByteBufferAllocator
+    ) -> ByteBuffer? {
+        let bytes = QUICConnectionUtilities.createStatelessResetPacket(
+            token: self.token(for: connectionID),
+            triggeringPacketLength: triggeringPacketLength
+        )
+        // No bytes means no reset could be built which is both a plausible QUIC packet (21 bytes
+        // at minimum) and smaller than the packet that triggered it.
+        guard !bytes.isEmpty else { return nil }
```
glbrntt: I should have a shortcut for this comment 🥲 – `guard !x else { ... }` is just a less readable version of `if x { ... }`

```suggestion
        if bytes.isEmpty { return nil }
```

Sources/NIOQUIC/SwiftNetwork/QUICStatelessResetTokenGenerator.swift
josephnoir: Good point! It takes a `DataProtocol`, so this should be possible.

josephnoir:
Thanks, George! I also switch the new test file to use Swift Testing.

Sources/NIOQUIC/SwiftNetwork/QUICStatelessResetTokenGenerator.swift
glbrntt: Related question that I really should've thought of at the same time: what's the return type of `HMAC<SHA256>.authenticationCode(for:using:)`? Can we avoid the `Array` creation on the next line as well?

josephnoir: We can indeed. I merged the two functions with an `InlineArray` as the intermediate storage.

Sources/NIOQUIC/SwiftNetwork/QUICStatelessResetTokenGenerator.swift
```diff
@@ -0,0 +1,91 @@
...
+    /// The stateless reset token for `connectionID`.
+    func token(for connectionID: QUICConnectionID) -> QUICStatelessResetToken {
+        let code = connectionID.withUnsafeBufferPointer {
+            HMAC<SHA256>.authenticationCode(for: $0, using: self.key)
+        }
+        let token = InlineArray<16, UInt8>(initializingWith: { outputSpan in
+            for elem in code.prefix(Self.tokenLength) {
+                outputSpan.append(elem)
+            }
+        })
```
glbrntt: Trailing closure syntax :)

```suggestion
        let token = InlineArray<16, UInt8> { outputSpan in
            for elem in code.prefix(Self.tokenLength) {
                outputSpan.append(elem)
            }
        }
```

glbrntt (approved):

PR #82, opened 2026-08-28
Avoid fabricating QUIC application error codes

6639835:
### Motivation

A NIO read-side close does not necessarily provide an application-defined QUIC error code. However, the current implementation sends `STOP_SENDING` even when no application error code is available, causing an error code of `0` to be fabricated.

Per RFC 9000, the error code carried by `STOP_SENDING` is an application protocol error code and therefore should come from the application rather than being synthesized by the transport.

This addresses #3.

### Modifications

* Skip sending `STOP_SENDING` when a NIO read-side close does not provide an application-defined error code.
* Continue sending `STOP_SENDING` when an explicit QUIC application error code is provided.
* Preserve explicitly supplied application error codes unchanged.
* Add regression coverage for both:

  * read-side closes without an application error code;
  * read-side closes with an explicit application error code.

### Result

NIOQUIC no longer fabricates application error code `0` for code-less read-side closes.

Applications that explicitly provide a QUIC application error code retain the existing behavior, with that code propagated unchanged.

### DIFF
--- a/Sources/NIOQUIC/SwiftNetwork/QUICApplicationAbortError.swift
+++ b/Sources/NIOQUIC/SwiftNetwork/QUICApplicationAbortError.swift
@@ -1,0 +1,43 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2026 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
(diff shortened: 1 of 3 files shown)


josephnoir:
Thank you for taking a look at this!

I don't think the solution can be to just skip sending `STOP_SENDING` as this can easily lead to an inconsistent state. The challenge is that SwiftNIO's directional close (`QUICChannelStreamHandler.close0`) does not have an argument for the application error code, and adding it is not a change that we want to make in SwiftNIO. We considered using a channel option for this with a default value set in the configuration, but have not landed on a design.

The stream state machine could also enforce this and only return `.closeRead` as an action when an application error code was provided.

PR #83, opened 2026-08-28
Wire-up stateless token responses

josephnoir:
### Motivation

The QUIC RFC defines stateless reset in Section 10.3. This PR adds support for handing out stateless reset tokens and sending our stateless resets in response to packet without local state. Handling incoming stateless resets will be implemented separately.

### Modifications

* A new configuration option to set the key for token generation.
* Instantiate a token generator from the configured key in the QUICHandler and pass it to QUIC connections.
* Respond to short-header packets addressed to unknown connections with a stateless reset packet.

### Results

Unknown connections can be reset across restarts.


josephnoir:
Major because it adds a new argument to the initializer of ~`QUICConfiguration`~ `QUICHandler`.

Sources/NIOQUIC/SwiftNetwork/QUICConfiguration.swift
```diff
@@ -114,6 +114,11 @@ public struct QUICConfiguration: Sendable {
     /// Maximum datagram frame size in bytes. Set to 0 to disable datagrams.
     /// Defaults to 65535 (the max value allowed) as recommended in RFC 9221.
     public var maxDatagramFrameSize: Int
+    /// The key for generating stateless reset tokens (RFC 9000 § 10.3.2).
+    ///
+    /// Must be at least 32 bytes. Defaults to `nil`, in which case a key will be generated.
+    /// This must be set to enable consistent stateless resets across restarts.
+    public var statelessResetKey: [UInt8]?
```
glbrntt: What happens if this isn't 32 bytes long? We should validate and crash real early.

It might also be worth introducing a nominal type for this that's more explicit about the `nil` case. The type could then be responsible for validation too. Rough shape (with names I haven't really thought about):

```swift
public struct StatelessResetKey {
  enum Wrapped {
    case autogenerate
    case userProvided([UInt8])
  }

  public static var autogenerate: Self { ... }

  // TODO: choose a better name
  public static func userProvided(_ bytes: [UInt8]) -> Self { 
    // validate length
  }
  
}

File: Sources/NIOQUIC/SwiftNetwork/QUICStatelessResetTokenGenerator.swift [hunk 2]
```diff
@@ -21,7 +21,50 @@ import NIOCore
...
+        var buffer = allocator.buffer(capacity: bytes.count)
+        buffer.writeBytes(bytes)
+        return buffer
+    }
+}
+
+/// A default implementation of ``QUICStatelessResetTokenGenerator`` that
+/// uses `HMAC<SHA256>` to generate tokens.
+@available(anyAppleOS 26, *)
+public struct HMACSHA256QUICStatelessResetTokenGenerator: QUICStatelessResetTokenGenerator {
```
glbrntt: Boy that sure is a real mouthful of a name. I wonder if we should just name it "DefaultStatelessResetTokenGenerator" and then have the docs explain.

Sources/NIOQUIC/SwiftNetwork/SwiftNetworkQUICConnection.swift
```diff
@@ -79,6 +79,9 @@ final class SwiftNetworkQUICConnection {
 
     private var streamOptions: QUICStreamProtocol.QUICStreamOptions
 
+    /// Derives the stateless reset tokens advertised with the connection IDs this connection issues.
+    private let quicStatelessResetTokenGenerator: any QUICStatelessResetTokenGenerator
```
glbrntt: `quic` seems redundant in the name here.

Sources/NIOQUIC/SwiftNetwork/SwiftNetworkQUICConnection.swift
```diff
@@ -187,6 +196,7 @@ final class SwiftNetworkQUICConnection {
     static func client(
         configuration: QUICConfiguration,
         sourceConnectionID: QUICConnectionID,
+        quicStatelessResetTokenGenerator: any QUICStatelessResetTokenGenerator,
```
glbrntt: same here (and in other places) re: `quic` prefix

Sources/NIOQUIC/QUICHandler.swift
```diff
@@ -454,6 +482,56 @@ public final class QUICHandler {
...
+    /// - Parameters:
+    ///   - header: The parsed header of the unroutable packet.
+    ///   - envelope: The datagram it arrived in.
+    private func trySendStatelessReset(
+        for header: QUICPacketHeader,
+        triggeredBy envelope: AddressedEnvelope<ByteBuffer>
+    ) {
+        self.eventLoop.assertInEventLoop()
+
+        // RFC 10.3 says that we may send stateless reset in response to long headers.
```
glbrntt: Presumably you meant this?

```suggestion
        // RFC 9000 § 10.3 says that we may send stateless reset in response to long headers.
```

Sources/NIOQUIC/SwiftNetwork/QUICStatelessResetTokenGenerator.swift
josephnoir: I'm okay either way. I was thinking of a hypothetical where the RFC recommendation changes and we want to change the default. We would need a new name (or change its implementation).

glbrntt: That's fair. It's just a lot of initialisms smooshed together making it hard to read. 

Here's one idea, why not nest the generator protocol in the `QUICStatelessResetToken` type, then the name of the concrete type becomes shorter and drops the QUIC prefix. In code it'd read as `QUICStatelessResetToken.HMACSHA256Generator`, something like this:

```swift
extension QUICStatelessResetToken {
  public protocol Generator {
    // ...
  }

  public struct HMACSHA256Generator: Generator {
    // ...
  }
}
```

Again, I don't mind too much since it's not a commonly use type (i.e. set and forget) but just food for thought. I'm happy to give with what you currently have too.

Sources/NIOQUIC/SwiftNetwork/QUICConfiguration.swift
josephnoir: The key is not longer a separate configuration. Going through [`.defaultWithUserProvidedKey(:)`](https://github.com/apple/swift-nio-quic/pull/83/changes#diff-a8cc84d3ef4708d4fd96d23e877db14d344d95003a66298cee3eb13874d02313R112) will directly instantiate the generator and check the key length. It is missing the precondition documentation though and I'll add that.

glbrntt: Sorry – I left that comment unpublished from the other day – didn't realize it was left in this batch!

Sources/NIOQUIC/SwiftNetwork/QUICStatelessResetTokenGenerator.swift
josephnoir: That would be easier to read. I like it!

glbrntt (approved):
LGTM, thanks @josephnoir!

PR #84, opened 2026-09-02
Align connection ID generator structure

josephnoir:
### Motivation

The recently introduced stateless reset token generators are extensions of the token type itself. This allows keeping the names a bit shorter and the coupling tighter. Both generator types should work in a similar way to make it easier to read and reason about the code.

### Modifications

* Drop "quic" prefix from `QUICHandler` connection ID generator arguments.
* Move the connection ID generator types into extensions of `QUICConnectionID`.

### Result

The connection ID and stateless reset token generators use the same patterns.


glbrntt (approved):

josephnoir:
Merging through expected failure of the api breakage check.

PR #85, opened 2026-09-02
Avoid crash on undersized datagrams

josephnoir:
### Motivation

Force unwrapping on uncontrolled input is dangerous.

### Modifications

Handle failures instead of force-unwrapping.

### Result

Robuster packet handling.

### DIFF
--- a/Sources/NIOQUIC/SwiftNetwork/QUICPacketHeader.swift
+++ b/Sources/NIOQUIC/SwiftNetwork/QUICPacketHeader.swift
@@ -257,6 +257,10 @@ extension NIOCore.ByteBuffer {
             packetType = .short
             let dcidLength = shortHeaderDCIDLength
 
+            guard localBuffer.readableBytes >= dcidLength else {
+                return nil
+            }
+
             let dcidWrittenBytes = destinationConnectionID.withUnsafeMutableBufferPointer {
                 destinationConnectionIDBytes in
                 localBuffer.readSlice(length: dcidLength)!.withUnsafeReadableBytes { pointer in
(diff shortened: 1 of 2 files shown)


Sources/NIOQUIC/SwiftNetwork/QUICPacketHeader.swift
```diff
@@ -257,9 +257,13 @@ extension NIOCore.ByteBuffer {
             packetType = .short
             let dcidLength = shortHeaderDCIDLength
 
+            guard let dcidSlice = localBuffer.readSlice(length: dcidLength) else {
+                return nil
```
glbrntt, line 261: presumably returning `nil` results in the packet just being dropped?

glbrntt (approved):

Sources/NIOQUIC/SwiftNetwork/QUICPacketHeader.swift
josephnoir: I was focused on the crash, but I think we need to revisit what happens when a packet fails to parse.

On a non-parsable packet `QUICHandler.channelRead(context:data:)`  throws `QUICError.quicPacketHeaderDecodingFailed`, which is caught and propagated with `context.fireErrorCaught(error)`. That seems a bit excessive.

Should this be more than a `log.debug`? Happy to address that in a separate PR.

glbrntt: Yeah, that's probably excessive. Firing an error will typically result in the channel being closed. Probably it's best to return `nil` and have the caller log. Fine for this to be a follow up.

PR #87, opened 2026-09-03
Less noise on packet parsing

josephnoir:
### Motivation

Parsing failures happen. Most likely not every packet sent to an open socket is part of an ongoing connection. Parsing failures should not be propagated as errors on the connection channel.

### Modification

* Do not throw on parsing failure and return nil instead.
* Log parsing failures instead of propagating them.
* Removed SwiftNIO QUIC's separate packet parsing.
* Fix header type detection, which assumed long headers.

### Results

Less noise. Avoid closing the channel when receiving unparsable packets.


Sources/NIOQUIC/SwiftNetwork/QUICPacketHeader.swift
```diff
@@ -83,7 +83,7 @@ struct QUICPacketHeader: Hashable, Sendable {
             self.init(base)
         }
 
-        init(firstByteAlreadyMasked: UInt8, version: Version) throws {
+        init(firstByteAlreadyMasked: UInt8, version: Version) {
```
glbrntt: Somewhat related to your PR: it looks like `firstByteAlreadyMasked` is masked to `0b11` so the cases below are fine _but_ the masking being applied elsewhere makes it harder to reason about as you need to look at the code used by each caller and it's easy to break which would potentially turn into an advisory. 

I think we should either do the masking here or add some defense in depth by failing the `init`/throwing or similar. We don't need to address this in this PR though.

Sources/NIOQUIC/SwiftNetwork/QUICPacketHeader.swift
```diff
@@ -185,8 +185,8 @@ extension NIOCore.ByteBuffer {
     /// - Returns: The parsed `QUICPacketHeader` or `nil` if the not enough bytes were readable.
     func parseQUICPacketHeader(
         destinationIDLength shortHeaderDCIDLength: Int
-    ) throws -> QUICPacketHeader? {
-        let routingHeader: QUICPacketHeader? = try self.withUnsafeReadableBytes { buffer in
+    ) -> QUICPacketHeader? {
```
glbrntt, line 203: The docs need updating here. The general NIO pattern is to do what the existing code does: throw on error, return `nil` on reading short. We could use a typed throws here and switch over the errors in the caller and decide what to do there instead of unconditionally returning `nil` on error.

Sources/NIOQUIC/SwiftNetwork/QUICPacketHeader.swift
josephnoir: Sounds like a good idea.

Sources/NIOQUIC/SwiftNetwork/QUICPacketHeader.swift
josephnoir: > The docs need updating here. 

True!

josephnoir:
> We could use a typed throws here and switch over the errors in the caller and decide what to do there instead of unconditionally returning nil on error.

Your comment inspired me to check if we could simply remove the packet parsing function implemented by SwiftNIO QUIC. And it is looking good. That change revealed some issues with tests and the packet type calculation.

Sources/NIOQUIC/SwiftNetwork/QUICPacketHeader.swift
josephnoir: I removed that initializer. It assumes long header packets and we want to know about short header packets as well.

Sources/NIOQUIC/SwiftNetwork/QUICPacketHeader.swift
```diff
@@ -83,13 +81,23 @@ struct QUICPacketHeader: Hashable, Sendable {
...
-        init(firstByteAlreadyMasked: UInt8, version: Version) throws {
+        /// Parse the packet type from the byte of the packet.
+        init(firstByte: UInt8, version: Version) {
+            // Check if the byte indicates a short header.
+            guard firstByte.indicatesLongHeader else {
+                self = .short
+                return
+            }
+
+            // Must be VN or a long header.
```
glbrntt: VN == version number? Helpful to link to or include snippets from the RFC for parsing.

glbrntt (approved):
LGTM!

josephnoir:
Expected API break because of changes to `QUICError`.

PR #88, opened 2026-09-04
Add paged slot storage for QUIC stream state

glbrntt:
Motivation:

Having a per-stream Channel is expensive unless the stream is long-lived. In order to get great performance we need an alternative API surface that is significantly cheaper.

The first phase of these changes will put together a table of per-stream state for a connection. The table has slots for transport state (i.e. the underlying QUIC stack) and application state (i.e. state stored on behalf of the consumer). Both states are `~Copyable`.

Each state store is backed by a `PagedBuffer` which allocates a buffer pointer for N values in pages. Pagres aren't deallocated until the buffer is destroyed, so the address of a value is stable throughout its lifetime.

Two stores sit on top of the buffer: `QUICStreamSlots` and another type for application state (not added in this PR). `QUICStreamSlots` provides a handle when a value is inserted which is invalidated when the value is removed (so a handle cannot access data it doesn't not own). The storage is dense: slots can be reused, and lookups are fast (as there is no hashing).

Modifications:

- Add `PagedBuffer`, discontiguous index-addressed storage whose pages are allocated once and never moved
- Add `Page` which describes how big a page is, and how an index into the paged buffer maps to a page and the position on that page.
- Add `QUICStreamHandle`, an index into the paged buffer, and a generation
- Add `QUICStreamSlotRecord`, recording the occupancy and generation of a slot
- Add `QUICStreamSlots`, which is the generation-checked slot storage backed by a `PagedBuffer`.

Result:

A few key pieces of infra are in place.

### DIFF
--- a/Sources/NIOQUIC/Collections/PagedBuffer.swift
+++ b/Sources/NIOQUIC/Collections/PagedBuffer.swift
@@ -1,0 +1,201 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2026 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// Discontiguous storage for values addressed by index.
+///
+/// The storage is made up of pages which begin at one slot and double in size up to 64. Pages are
+/// allocated once and never moved, so the address of each value is stable until the buffer is
+/// destroyed and can be accessed via a pointer. By itself this construct is _not_ safe; callers
+/// must take care to ensure that pointers do not outlive the buffer.
+///
+/// Slots are handed out densely from zero: the buffer grows by one slot at a time and never has a
+/// hole, so its ``count`` is the next index it will hand out.
+@usableFromInline
+struct PagedBuffer<Value: ~Copyable>: ~Copyable {
+    /// The allocated pages.
+    @usableFromInline
+    var _pages: [UnsafeMutablePointer<Value>]
+
+    /// The number of slots handed out.
+    @usableFromInline
+    var _count: Int
+
+    @inlinable
+    init() {
+        self._pages = []
+        self._count = 0
+    }
+
+    deinit {
+        for page in self._pages {
+            page.deallocate()
+        }
+    }
+
+    /// The number of slots handed out, which is also the index the next ``append()`` returns.
+    @inlinable
+    var count: Int {
+        self._count
+    }
+
+    /// The number of pages allocated so far.
+    @inlinable
+    var pageCount: Int {
+        self._pages.count
+    }
+
+    /// The base pointer for a given `page`.
+    ///
+    /// - Precondition: `page` must exist.
+    @inlinable
+    func basePointer(ofPage page: Int) -> UnsafeMutablePointer<Value> {
+        self._pages[page]
+    }
+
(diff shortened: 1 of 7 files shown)


Sources/NIOQUIC/Collections/PagedBuffer.swift
```diff
@@ -0,0 +1,202 @@
...
+    ///
+    /// The slot's storage is allocated but not initialized. It is the caller's responsibility to
+    /// initialize it. If `Value` is an `Optional` (or expressible by `nil`) then you can grow the
+    /// buffer with `nil`-initialized storage using ``appendNil()``.
+    @inlinable
+    @discardableResult
+    mutating func append() -> UnsafeMutablePointer<Value> {
+        let position = Page.position(of: self._count)
+
+        // Slots are dense, so the new one lands either on the last page or on the very next one.
```
josephnoir, line 77: Maybe add an `assert(position.pageIndex >= self.PageCount - 1)`? Although the tests on this look very thorough.

Sources/NIOQUIC/Collections/PagedBuffer.swift
```diff
@@ -0,0 +1,202 @@
...
+    }
+}
+
+@usableFromInline
+enum Page {
+    /// The capacity every page has once doubling stops.
+    @usableFromInline
+    static var maxPageCapacity: Int { 64 }
+
+    /// The first page which holds ``maxPageCapacity`` slots.
```
josephnoir:
```suggestion
    /// The index of the first page which holds ``maxPageCapacity`` slots.
```

Sources/NIOQUIC/Streams/Transport/QUICStreamSlots.swift
```diff
@@ -0,0 +1,179 @@
...
+@usableFromInline
+struct QUICStreamSlots<Value: ~Copyable>: ~Copyable {
+    @usableFromInline
+    var _storage: PagedBuffer<Value>
+
+    /// A record of each allocated slot, indexed by the `Index` of a `QUICStreamHandle`.
+    @usableFromInline
+    var _slots: [QUICStreamSlotRecord]
+
+    /// The indexes of the vacant slots.
```
josephnoir:
```suggestion
    /// The indices of the vacant slots.
```

Think this is the one we have used so far.

Sources/NIOQUIC/Streams/QUICStreamHandle.swift
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        /// The first slot a store hands out.
+        @inlinable
+        static var first: Index {
+            Index(0)
+        }
+
+        @inlinable
+        mutating func formNext() {
```
josephnoir: How about `advance`? "form" sounds like it produces something.

josephnoir (approved):
Looks great! I have a few suggestions that you can pick up if you like.

Sources/NIOQUIC/Streams/QUICStreamHandle.swift
glbrntt: `formIndex` is used on a number Swift collections, but typically in a slightly different way in that the index is passed `inout` to the collection which advances it:

```swift
var array = [1, 2, 3]
var index = array.startIndex
array.formIndex(after: &index)
```

That's the reason I picked it but I'm equally happy with `advance`. Let me know your preference!

Sources/NIOQUIC/Collections/PagedBuffer.swift
glbrntt: Yup, nice idea. I added an `else` branch which asserts exactly `pageCount - 1` to tighten it up some more.

Sources/NIOQUIC/Streams/QUICStreamHandle.swift
glbrntt: I went with `advance`, internal only and its clear.

PR #89, opened 2026-09-07
Add QUICStreamCore for one stream's transport state

glbrntt:
Motivation:

The stream table needs a type per-stream which owns various bits of state.

Note that this mostly wraps other things so testing it in isolation isn't really worth it. It'll be tested indirectly in a later change. For now the API is also unsued.

Modifications:

- Add QUICStreamCore
- Add QUICStreamReadOutcome, what a read hands back.
- Add QUICStreamStateMachine.isSendFinished, which asks whether a FIN or RESET_STREAM has gone out without waiting for the peer to acknowledge it.

Result:

One stream's transport state has a home

### DIFF
--- a/Sources/NIOQUIC/Streams/QUICStreamReadOutcome.swift
+++ b/Sources/NIOQUIC/Streams/QUICStreamReadOutcome.swift
@@ -1,0 +1,23 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2026 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// What a read took from a stream.
+public enum QUICStreamReadOutcome: Hashable, Sendable {
+    /// The number of bytes read.
+    case read(Int)
+    /// The number of bytes read, after which the stream was closed.
+    case endOfStream(Int)
+    /// No bytes were available to read.
+    case nothingAvailable
+}
--- a/Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
+++ b/Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
@@ -203,7 +203,7 @@ extension QUICStreamCore {
     ///   - minContiguous: The shortest run of contiguous bytes to hand to `body`. Data is
     ///     coalesced to reach it; `1` never coalesces. A run can still be shorter than
     ///     `minContiguous` if `maxBytes` is smaller.
-    ///   - body: Provided a contiguous block of bytes received from the remote peet, returning how
+    ///   - body: Provided a contiguous block of bytes received from the remote peer, returning how
     ///     many of them it consumed. May be called more then once for each call to `read`.
     /// - Returns: The outcome of the read.
     @usableFromInline
@@ -576,6 +576,7 @@ extension QUICStreamCore {
                     refusalReason = "RESET_STREAM has already been sent"
                 }
             } catch {
+                assert(self.pendingWrities.isEmpty)
                 throw NetworkError(streamStateViolation: "\(error)", operation: "sendFin")
             }
         }
--- a/Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
+++ b/Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
@@ -1,0 +1,798 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2026 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import DequeModule
+import NIOCore
+import NIOQUICHelpers
+@_spi(Essentials) @_spi(ProtocolProvider) import SwiftNetwork
+
+/// The transport state for a stream.
+@available(anyAppleOS 26, *)
+@usableFromInline
+struct QUICStreamCore: ~Copyable {
+    /// The ID of the stream assigned by `SwiftNetwork`, `nil` if it hasn't been assigned yet
+    /// (e.g. is locally opened).
+    @usableFromInline
+    private(set) var id: QUICStreamID?
+
+    /// The state of a stream.
+    private var state: QUICStreamStateMachine
+
+    /// Set when a frame had the FIN flag set, cleared when the FIN is handed to the state machine.
+    ///
+    /// This is necessary so that the state machine doesn't receive the FIN before the consumer has
+    /// received all inbound data (as each read must also go through the state machine).
+    private var finPending: Bool
+
+    /// Writes that haven't yet been emitted to the stack.
+    private var pendingWrites: FrameArray
+
+    /// Data received from `SwiftNetwork` which the application hasn't consumed yet.
+    private var undeliveredReads: UniqueDeque<Frame>
+
+    /// The handle used to talk to the SwiftNetwork stack.
+    private var handle: SwiftNetworkStreamHandle
+
+    /// The reference every linkage call is made "from".
+    ///
+    /// Holding it creates a retain cycle (the ref, holds the container, holds this stream) which
+    /// is broken in ``close(error:)``.
+    private var reference: ProtocolInstanceReference
+
+    /// Temporary space for coalescing into when the consumer asks for reads to be coalesced and
+    /// the leading frame doesn't contain enough bytes.
+    private var coalescedBytes: [UInt8]
+
+    init(id: QUICStreamID?, reference: ProtocolInstanceReference, linkage: OutboundStreamLinkage) {
+        self.id = id
+        self.state = QUICStreamStateMachine()
+        self.handle = SwiftNetworkStreamHandle(linkage: linkage)
+        self.reference = reference
+        self.pendingWrites = FrameArray()
+        self.undeliveredReads = UniqueDeque()
+        self.finPending = false
+        self.coalescedBytes = []
+    }
+
+    /// Creates a stream core which is not yet wired to the stack.
+    ///
+    /// - Parameter id: The stream ID, if the stack has already assigned one.
+    init(id: QUICStreamID?) {
+        self.id = id
+        self.state = QUICStreamStateMachine()
+        self.handle = SwiftNetworkStreamHandle()
+        self.reference = ProtocolInstanceReference()
+        self.pendingWrites = FrameArray()
+        self.undeliveredReads = UniqueDeque()
+        self.finPending = false
+        self.coalescedBytes = []
+    }
+
+    /// Attaches this stream core to the stack.
+    ///
+    /// - Parameters:
+    ///   - reference: The reference the stack routes this stream's events back through, which is
+    ///     also what every linkage call is made "from".
+    ///   - linkage: The flow the stack attached for this stream.
+    mutating func attach(reference: ProtocolInstanceReference, linkage: OutboundStreamLinkage) {
+        self.reference = reference
+        self.handle = SwiftNetworkStreamHandle(linkage: linkage)
+    }
+
+    /// The stack's metadata for this stream, or `nil` if it is detached.
+    func metadata() -> ProtocolMetadata<QUICProtocol>? {
+        switch self.handle.invokeGetMetadata() {
+        case .proceed(let linkage):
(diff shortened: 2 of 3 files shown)


Sources/NIOQUIC/SwiftNetwork/QUICStreamStateMachine.swift
```diff
@@ -222,6 +222,30 @@ struct QUICStreamStateMachine: ~Copyable {
         }
     }
 
+    /// Returns `true` once a FIN or RESET\_STREAM has been sent, the stream has no send side, or
+    /// the stream is closed.
+    ///
+    /// Unlike ``isWriteClosed`` this doesn't wait for the peer to acknowledge what was sent.
+    var isSendFinished: Bool {
```
josephnoir, line 229: The naming would be more consistent with `isWriteClosed` if we named it `isWriteFinished`. Then again, the underlying state is called `sendState`. Maybe we should rename `isWriteClosed` to `isSendClosed`.

Sources/NIOQUIC/Streams/QUICStreamReadOutcome.swift
```diff
@@ -0,0 +1,23 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// What a read took from a stream.
+public enum QUICStreamReadOutcome: Hashable, Sendable {
+    /// The number of bytes read.
+    case read(Int)
+    /// The number of bytes read, after which the stream was closed.
+    case endOfStream(Int)
```
josephnoir:
```suggestion
    case endOfStream(readBytes: Int)
```

Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
```diff
@@ -0,0 +1,798 @@
...
+    /// Attaches this stream core to the stack.
+    ///
+    /// - Parameters:
+    ///   - reference: The reference the stack routes this stream's events back through, which is
+    ///     also what every linkage call is made "from".
+    ///   - linkage: The flow the stack attached for this stream.
+    mutating func attach(reference: ProtocolInstanceReference, linkage: OutboundStreamLinkage) {
+        self.reference = reference
+        self.handle = SwiftNetworkStreamHandle(linkage: linkage)
+    }
```
josephnoir, line 100: Is it okay for this to happen more than once?

Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
```diff
@@ -0,0 +1,798 @@
...
+            return linkage.invokeGetMetadata(self.reference) as ProtocolMetadata<QUICProtocol>?
+        case .ignore:
+            return nil
+        }
+    }
+
+    /// Whether more data can be written to the stream.
+    @usableFromInline
+    var isSendOpen: Bool {
+        !self.state.isSendFinished
```
josephnoir: Does this include `self.state.isWriteClosed`?

Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
```diff
@@ -0,0 +1,798 @@
...
+        switch action {
+        case .closeStream:
+            ()
+        case .surfaceReset:
+            ()
+        case .doNothing(.alreadyFullyReceived):
+            ()  // The peer already finished cleanly; the reset has nothing left to abandon.
+        case .doNothing(.alreadyReset):
+            ()  // RESET_STREAM is not acted on twice.
+        }
```
josephnoir, line 166: Looks like this function could be a single line. More serious: The actions that imply a consequence don't have one, can we document why nothing needs to be done?

Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
```diff
@@ -0,0 +1,798 @@
...
+            action = try self.state.receiveStopSending(applicationErrorCode: code)
+        } catch {
+            return  // Not connected, or a direction with no send side.
+        }
+
+        switch action {
+        case .sendReset:
+            self.abortSendSideForStopSending(code: code)
+        case .sendResetAndCloseStream:
+            self.abortSendSideForStopSending(code: code)
```
josephnoir: Why do they do the same even though the actions have a different name?

Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
```diff
@@ -0,0 +1,798 @@
...
+    /// `body` returns the number of bytes that it consumed from the span. Bytes which weren't read
+    /// are held until the next read. This function will return `endOfStream` if the stream is half
+    /// closed.
+    ///
+    /// - Parameters:
+    ///   - maxBytes: The maximum number of bytes to read.
+    ///   - minContiguous: The shortest run of contiguous bytes to hand to `body`. Data is
+    ///     coalesced to reach it; `1` never coalesces. A run can still be shorter than
+    ///     `minContiguous` if `maxBytes` is smaller.
+    ///   - body: Provided a contiguous block of bytes received from the remote peet, returning how
```
josephnoir:
```suggestion
    ///   - body: Provided a contiguous block of bytes received from the remote peer, returning how
```

Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
```diff
@@ -0,0 +1,798 @@
...
+                break
+            }
+
+            let effectiveReadLength = min(minContiguous, available)
+            let firstLength = self.undeliveredReads[0].unclaimedLength
+            let consumed: Int
+
+            if firstLength >= effectiveReadLength {
+                if let bytes = self.undeliveredReads[0].bytes {
+                    consumed = min(max(body(bytes), 0), firstLength)
```
josephnoir: I'd prefer if we made a divergence between our expectation and the result of `body(...)` visible. 

```suggestion
                    let consumedByBody = min(max(body(bytes), 0), firstLength)
                    assert((0...firstLength).contains(consumedByBody))
                    consumed = consumedByBody
```

We can keep the `min(max(...))` if it helps avoid unexpected problems.

Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
```diff
@@ -0,0 +1,798 @@
...
+
+            if let source = self.undeliveredReads[index].span {
+                let slice = source.extracting(0..<min(source.count, wanted))
+                slice.withUnsafeBufferPointer { pointer in
+                    self.coalescedBytes.append(contentsOf: pointer)
+                }
+            }
+        }
+
+        let consumed = min(max(body(self.coalescedBytes.span.bytes), 0), self.coalescedBytes.count)
```
josephnoir: Same question about the return value of `body`.

Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
```diff
@@ -0,0 +1,798 @@
...
+
+// MARK: - Writing
+
+@available(anyAppleOS 26, *)
+extension QUICStreamCore {
+    /// How many bytes the stack will take right now, which is advisory: a write past it is buffered
+    /// rather than refused.
+    ///
+    /// Note that this returns `0` until the stack has assigned the stream its ID, which is not the
+    /// same as no room: a write before then is legal and is buffered.
```
josephnoir, line 486: Is that buffer limited?

Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
```diff
@@ -0,0 +1,798 @@
...
+    ///   - body: Handed `count` writable bytes, returning how many it wrote.
+    @usableFromInline
+    mutating func withWritableSpan(count: Int, _ body: (inout MutableRawSpan) -> Int) {
+        var frame = Frame(allocatingCustomFinalizerBufferOfSize: count)
+        var written = 0
+
+        if var span = frame.mutableSpan {
+            var bytes = span.mutableBytes
+            let wrote = body(&bytes)
+            written = min(max(wrote, 0), count)
```
josephnoir, line 517: Assert?

Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
```diff
@@ -0,0 +1,798 @@
...
+        self.pendingWrites.add(frame: frame)
+    }
+
+    /// Queues a copy of `buffer`.
+    ///
+    /// - Parameter buffer: The bytes to queue.
+    @usableFromInline
+    mutating func write(_ buffer: ByteBuffer) {
+        if buffer.readableBytes != 0 {
+            self.pendingWrites.add(frame: Frame(copyBuffer: buffer.readableBytesUInt8Span))
```
josephnoir: Could we get around a copy with this API: https://github.com/apple/swift-nio-quic/blob/main/Sources/NIOQUIC/SwiftNetwork/SwiftNetworkQUICConnection.swift#L929:

```swift
packet.withUnsafeMutableReadableBytesWithStorageManagement2 { buffer, owner in
    self.inputPacketQueue.add(frame: Frame(customBuffer: buffer, owner: owner))
}
```

Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
```diff
@@ -0,0 +1,798 @@
...
+                switch try self.state.sendFin() {
+                case .sendFin:
+                    carriesFin = true
+                case .ignore(.alreadyFinished):
+                    refusalReason = "a FIN has already been sent"
+                case .ignore(.streamReset):
+                    refusalReason = "RESET_STREAM has already been sent"
+                }
+            } catch {
+                throw NetworkError(streamStateViolation: "\(error)", operation: "sendFin")
```
josephnoir, line 592:
```suggestion
            } catch {
                assert(self.pendingWrities.isEmpty)
                throw NetworkError(streamStateViolation: "\(error)", operation: "sendFin")
```

Nit: Hopefully not needed, but I'd rather hit the assert than go looking for a path where we might not have finalized frames.

josephnoir (commented):
Looks great overall! Loads of bytes get pushed around here and I'm looking forward to the layer that adds tests for it. I have a few comments, mostly regrarding a few more checks.

Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
josephnoir: Looks like you addressed this already. Thanks!

Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
glbrntt: Hmm, probably not? I don't think it is called more than once but I'll add an `assert` to be sure.

Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
glbrntt: True, good idea!

Sources/NIOQUIC/SwiftNetwork/QUICStreamStateMachine.swift
glbrntt: Yeah, I don't mind. I can rename `isWriteClosed` to is `isSendClosed` if you like? I'll do it in a separate PR though if that's alright?

josephnoir: A separate sounds good. No reason to do it here. Thanks!

Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
glbrntt: Great question 😄

The state machine is really designed for `QUICChannelStreamHandler` but a lot of its logic _is_ useful here (and subsequent PRs). In this stream model we receive events from SwiftNetwork but don't want to act on them just yet, we want to do so at the end of an event-loop tick. We still need to transition the state machine, but observe side effects later on.

Once this becomes the canonical path I suspect we'll be able to simplify the state machine a lot.

Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
glbrntt: Yes, good idea!

Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
josephnoir: It has also already been tested quite a bit, so it's definitely good to reuse it.

> Once this becomes the canonical path I suspect we'll be able to simplify the state machine a lot

I was hoping to read that. 😁

Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
glbrntt: No, but it's only your own process that you can bloat the memory of. This is much the same as ignoring a writability notification in swift-nio: you can only bloat your own buffers.

josephnoir: Fair.

josephnoir (approved):
Thanks!

PR #90, opened 2026-09-07
Rename isWriteClosed to isSendClosed

glbrntt:
Motivation:

The state machine tracks send state, so this is a little more consistent. Discussed in #89.

Modifications:

Rename

Result:

More consistency


josephnoir (approved):
Thank you so much!

PR #91, opened 2026-09-07
Add per-stream storage for a consumer's state

glbrntt:
Motivation:

The connection table will need to store the per stream consumer state. This will be provided by the `QUICStreamConsumer` and stored in `QUICStreamConsumerStates`.

Modifications:

- Add `QUICStreamConsumer`, this will be the protocol that users implement. At the moment it only provides the `associatedtype` for the state it provdes and isn't actually used yet.
- Make the `nil`-handling extension on `PagedBuffer` handle `~Copyable` values (I missed doing that previously)
- Add "Lifetimes" to the manifest (not user here, but makes the next PR a little smaller)

Result:

Not much changes beyond making the next PR a little smaller


josephnoir (approved):
👍

PR #92, opened 2026-09-08
Add QUICStreamTable

glbrntt:
Motivation:

Many of the core pieces are in place for lightweight streams now so we can start putting them together.

Modifications:

- Add `QUICStreamEvents` which is a bitset of events that have happened to a stream since its last visit. Users can look at this to know, for example, whether it's worth attempting to pull bytes out of the stack.
- Add 'QUICStream', the public facing part of a stream presented to a consumer with various operations on it.
- Add 'QUICStreamTransportState', the state of the transport side of a stream as held the stream table
- Add 'QUICStreamTable' which pieces the above any a number of other bits together. It allows new streams to be inserted into storage and accessed via the handle associated with the storage. It also provides a 'withStream' function which provides scoped access to a stream and its underling storage.

Result:

Things are coming together. This is still not reachable.

### DIFF
--- a/Sources/NIOQUIC/Streams/QUICStream.swift
+++ b/Sources/NIOQUIC/Streams/QUICStream.swift
@@ -1,0 +1,192 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2026 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+import NIOQUICHelpers
+
+/// A view over a QUIC stream as presented to a ``QUICStreamConsumer``.
+@available(anyAppleOS 26, *)
+public struct QUICStream<Consumer: QUICStreamConsumer & ~Copyable>: ~Copyable, ~Escapable {
+    /// The table this stream's slot belongs to.
+    @usableFromInline
+    let table: QUICStreamTable<Consumer>
+
+    /// The transport state for the stream.
+    @usableFromInline
+    let transport: UnsafeMutablePointer<QUICStreamTransportState>
+
+    /// An opaque handle identifying this stream.
+    ///
+    /// The handle is invalidated when the stream is closed.
+    public let handle: QUICStreamHandle
+
+    // 'immortal' because the lifetime really depends on the lifetime of the 'transport' pointer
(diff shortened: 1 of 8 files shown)


Sources/NIOQUIC/Streams/Transport/QUICStreamTable.swift
```diff
@@ -0,0 +1,261 @@
...
+    @inlinable
+    var hasPendingWork: Bool {
+        !self._ready.isEmpty
+    }
+}
+
+// MARK: - Bookkeeping
+
+@available(anyAppleOS 26, *)
+extension QUICStreamTable where Consumer: ~Copyable {
```
josephnoir: Since `Consumer` has the requirement `QUICStreamConsumer & ~Copyable`, this is more of a reminder?

Sources/NIOQUIC/Streams/QUICStream.swift
```diff
@@ -0,0 +1,192 @@
...
+        minContiguousBytes: Int = 1,
+        _ body: (_ span: borrowing RawSpan) -> Int
+    ) -> QUICStreamReadOutcome {
+        self.transport.pointee.core.read(maxBytes: maxBytes, minContiguous: minContiguousBytes, body)
+    }
+
+    /// Appends the stream's inbound bytes to `buffer`.
+    ///
+    /// - Parameters:
+    ///   - maxBytes: The most bytes to take out of the stack.
```
josephnoir: Is this also a hint?

Sources/NIOQUIC/Streams/QUICStream.swift
```diff
@@ -0,0 +1,192 @@
...
+    ///
+    /// - Parameters:
+    ///   - maxBytes: The maximum number of bytes to read from the network stack in this call. Note
+    ///     that this isn't a limit on the total number of bytes passed to the span passed to
+    ///     `body`; the actual byte count may differ so you should consider this a hint.
+    ///   - minContiguousBytes: The shortest run of contiguous bytes to hand to `body`. Data is
+    ///     coalesced to reach it (which may incur additional copies and allocations); `1` never
+    ///     coalesces (and therefore does not incur additional copies and allocations).
+    ///   - body: Called with a chunk of contiguous bytes, returning how many of them it consumed.
+    ///     May be called multiple times.
```
josephnoir: Two comments on this:
1. Since `maxBytes` is more of a hint, maybe we can just get rid of it? As it stands the caller has to do all the bookkeeping, so it does not save you a lot.
2. `minContiguousBytes`: Just double checking that this means you will never get less bytes in a call. I remember there being a dependency on `maxBytes` that is not documented here.

Sources/NIOQUIC/Streams/QUICStream.swift
josephnoir: Same comment. If we keep it, we should clarify if this is a hint as well.

josephnoir (approved):
Nice one!

PR #93, opened 2026-09-08
Add a quic stream proxy

glbrntt:
Motivation:

At the moment SwiftNetwork has no way of accessing the stream table or storing events for a stream.

This change takes a little explaining... In SwiftNetwork a `ProtocolInstanceContainer` can store and service one or many `ProtocolInstance`s. A `ProtocolInstanceReference` is the reference to a single instance within a container and can use an `Int` to identify a single instance in a container (if the container stores multiple instances).

This change introduce a few things: `QUICStreamTable` becomes a `ProtocolInstanceContainer`, `QUICStreamProxy` is added and becomes a `ProtocolInstance`.

This leaves `ProtocolInstanceReference`: our idea of identity for a stream is a `QUICStreamHandle` which is currently 64-bits wide. This is an issue because `Int` isn't always 64-bits.

Modifications:

- Add a nominal type for the generation in a `QUICStreamHandle`
- Change the generation and slot index to be backed by half of a `UInt` so that handle can be packed into an `Int`. This is a bit limiting on 32-bit platforms but probably fine. (SwiftNetwork has plans to change this mechanism soon anyway so it'll likely just go away as an issue anyway.)
- Add `QUICStreamProxy`, effectively a handle and a table ref.
- Make `QUICStreamTable` a `ProtocolInstanceContainer`.

Result:

SwiftNetwork can reach the stream table (if we wire it up)

### DIFF
--- a/Sources/NIOQUIC/Streams/Transport/QUICStreamProxy.swift
+++ b/Sources/NIOQUIC/Streams/Transport/QUICStreamProxy.swift
@@ -1,0 +1,154 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2026 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOQUICHelpers
+@_spi(Essentials) @_spi(ProtocolProvider) import SwiftNetwork
+
+#if canImport(Darwin)
+import Darwin
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
+#endif
+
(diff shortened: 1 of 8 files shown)


Sources/NIOQUIC/Streams/Transport/QUICStreamProxy.swift
```diff
@@ -0,0 +1,154 @@
...
+
+    var reference: ProtocolInstanceReference {
+        self.table.reference(for: self.handle)
+    }
+
+    /// The slot's event manager, or the table's spare if the slot has been recycled.
+    var eventManager: ProtocolEventManager {
+        // The event manager in each `else` branch _should_ be unreachable: a reference is only
+        // ever built for a live handle, and all stack callbacks go via the table which drops
+        // handles that no longer resolve.
```
josephnoir, line 50: Is the `_should_` strong enough for `assertionFailure("ooops")`?

josephnoir (approved):
I like it.

Sources/NIOQUIC/Streams/Transport/QUICStreamProxy.swift
glbrntt: Yeah, I think so.

PR #94, opened 2026-09-09
Add the stream iterator, visit, and consumer API

glbrntt:
Motivation:

We have a stream table that can be hooked up to SwiftNetwork, but the consumer API doesn't exist. This PR adds it and wires it up to the table.

There are a few new parts here:
- `QUICStreamIterator`, an iterator over the ready streams in the table
- `QUICStreamVisit`, a single visit to a ready stream produced by the iterator, this is the API that users interact with in the consumer
- `QUICStreams` a handle for getting access to streams out-of-band (e.g. outside of a read loop)

Modifications:

- Add the types listed above
- Wire them up to the stream table
- More tests driven by the table

Result:

Aside from being wired into the actual stack, this is more or less the stream part done.

### DIFF
--- a/Sources/NIOQUIC/Streams/QUICStreamIterator.swift
+++ b/Sources/NIOQUIC/Streams/QUICStreamIterator.swift
@@ -1,0 +1,101 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2026 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOQUICHelpers
+
+/// An iterator for streams to visit.
+///
+/// Consuming a stream from the iterator means its events are considered consumer and won't be
+/// presented again. If a stream _isn't_ consumed by the iterator then it will be included in the
+/// next visit.
+@available(anyAppleOS 26, *)
+public struct QUICStreamIterator<Consumer: QUICStreamConsumer & ~Copyable>: ~Copyable, ~Escapable {
+    /// The handle and events that the last call to `next()` returned.
+    @usableFromInline
+    struct Pending {
+        @usableFromInline
+        var handle: QUICStreamHandle
+        @usableFromInline
+        var presented: QUICStreamEvents
+
+        @inlinable
+        init(handle: QUICStreamHandle, presented: QUICStreamEvents) {
+            self.handle = handle
+            self.presented = presented
+        }
+    }
+
+    @usableFromInline
+    let _table: QUICStreamTable<Consumer>
+
+    /// The handle and events which are currently being visited.
+    @usableFromInline
+    var _pending: Pending?
+
+    @inlinable
+    @_lifetime(immortal)
+    init(table: QUICStreamTable<Consumer>) {
+        self._table = table
+        self._pending = nil
+    }
+
+    deinit {
+        // Finish the last visit (in case the iterator wasn't completely consumed).
+        if let pending = self._pending {
+            self._table.finishVisit(pending.handle, presented: pending.presented)
+        }
+    }
+
+    /// The next stream to service, or `nil` if there are none left.
+    ///
+    /// - Returns: A stream visit.
+    @inlinable
+    @_lifetime(&self)
+    public mutating func next() -> QUICStreamVisit<Consumer>? {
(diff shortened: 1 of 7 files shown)


Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
```diff
@@ -265,9 +271,21 @@ extension QUICStreamCore {
...
+            self._needsReadVisit = true
+        }
+
         return outcome
     }
 
+    /// Whether the stream should be revisited in the next tick because it has unread data.
+    @inlinable
+    mutating func needsReadVisit() -> Bool {
+        defer { self._needsReadVisit = false }
```
josephnoir: Checking this twice delivers a different result even if nothing changed. Can we set `self._needsReadVisit = false` at the beginning of `read(minContiguous:_:)`?

Sources/NIOQUIC/Streams/Transport/QUICStreamTable.swift
```diff
@@ -276,6 +283,131 @@ extension QUICStreamTable where Consumer: ~Copyable {
...
+            state: UnsafeMutablePointer<Consumer.StreamState?>,
+            events: QUICStreamEvents
+        ) {
+            self.transport = transport
+            self.state = state
+            self.events = events
+        }
+    }
+
+    /// Returns the state of a ready slot prior to a visit, or `nil` if the handle is invalid.
```
josephnoir: This does not just return the state, maybe just the ready slot? What does the "prior to a visit" imply; is it a restriction on when this must be called?

Sources/NIOQUIC/Streams/QUICStreamIterator.swift
```diff
@@ -0,0 +1,101 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOQUICHelpers
+
+/// An iterator for streams to visit.
+///
+/// Consuming a stream from the iterator means its events are considered consumer and won't be
```
josephnoir:
```suggestion
/// Consuming a stream from the iterator means its events are considered consumed and won't be
```

glbrntt:
API breakage is expected and not an issue because `QUICStreamConsumer` hasn't been shipped yet nor is it used anywhere.

```
  💔 API breakage: func QUICStreamConsumer.makeStreamState(_:) has been added as a protocol requirement
  💔 API breakage: func QUICStreamConsumer.processStreams(_:) has been added as a protocol requirement
```

Sources/NIOQUIC/Streams/QUICStreamVisit.swift
```diff
@@ -0,0 +1,174 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOQUICHelpers
+
+@available(anyAppleOS 26, *)
+public struct QUICStreamVisit<Consumer: QUICStreamConsumer & ~Copyable>: ~Copyable, ~Escapable {
```
josephnoir, line 26: Since this is a public type it deserves a short description.

Sources/NIOQUIC/Streams/QUICStreamVisit.swift
```diff
@@ -0,0 +1,174 @@
...
+                table: self._table,
+                transport: self._transport,
+                handle: self.handle
+            )
+
+            yield &stream
+        }
+    }
+
+    /// Access to the other streams on this connection.
```
josephnoir: Very nitty nit: Other or `all`?

Tests/NIOQUICTests/Streams/Transport/QUICStreamTableTests.swift
```diff
@@ -201,4 +270,249 @@ struct QUICStreamTableTests {
...
+        table.closeAll(error: Boom(), disconnect: nil, into: &consumer)
+
+        #expect(consumer.sightings.count == 3)
+        // Second is enqueued first (when marked as ready), first and third are added to the
+        // queue by 'closeAll'.
+        #expect(consumer.sightings.map { $0.handle } == [second, first, third])
+        #expect(consumer.sightings.allSatisfy { $0.events.contains(.closed) })
+        #expect(table.count == 0)
+
+        table.drain(into: &consumer)
```
josephnoir: Why do we need this one?

Sources/NIOQUIC/Streams/Transport/QUICStreamTable.swift
glbrntt: This is just called by the iterator before creating the visit, I'll document the intended caller. Good point re: state, "state" encompasses more than `state` but the comment is somewhat ambiguous.

josephnoir (approved):
Nice, thanks George!

PR #95, opened 2026-09-09
Add a cheaper network scheduler

glbrntt:
Motivation:

The existing scheduler we plug in to SwiftNetwork is somewhat expensive as its based on scheduled tasks. NIO has a newer cheaper construct: `NIOScheduledCallback`.

Also, cancelling scheduled tasks is O(log N) where N is the number of scheduled tasks: this can get expensive when dealing with thousands of connections.

Modifications:

- Delete `QUICChannelEventLoop` (it's not an event loop) and replace it with `EventLoopBackedScheduler` which is a NetworkContext.Scheduler backed by an event loop
- The new scheduler uses `NIOScheduledCallback` and only cancels tasks when they are rescheduled for an earlier date. For tasks scheduled at a later date just let them run and then reschedule them for the newer date.

Result:

- Scheduling is cheaper (moves some QUIC benchmarks by a few percent)
- Scheduling has tests


Sources/NIOQUIC/SwiftNetwork/EventLoopBackedScheduler.swift
```diff
@@ -0,0 +1,181 @@
...
+        reference: SwiftNetwork.TimerReference
+    ) -> Bool {
+        wakeup.deadline = deadline
+        wakeup.task = task
+
+        let keepWakeup: Bool
+
+        // Only cancel the scheduled task if the new deadline is earlier. If the new deadline is
+        // the same as or later than the scheduled deadline then let the original callback fire as
+        // cancellation can be expsneive (it's O(log N) where N is the number of tasks currently
```
josephnoir:
```suggestion
        // cancellation can be expensive (it's O(log N) where N is the number of tasks currently
```

josephnoir (approved):

PR #99, opened 2026-09-10
Open streams from a stream table

glbrntt:
Motivation:

There's currently no API for opening streams via the stream table, only for accepting inbound streams.

Modifications:

- Add 'QUICStreamOpener' which holds what the bits required to attach an outbound flow.
- Add 'QUICStreamTable.open(_:state:)', which validates the type of stream being opened against the role of the local peer and drives the machinery to open a stream.
- Add 'QUICStreams.open(_:state:)' which is the public API for opening streams.

Result:

Streams can be opened

### DIFF
--- a/Sources/NIOQUIC/Streams/Transport/QUICStreamOpener.swift
+++ b/Sources/NIOQUIC/Streams/Transport/QUICStreamOpener.swift
@@ -23,9 +23,7 @@ struct QUICStreamOpener {
     /// The stack's connection instance.
     let connection: ProtocolInstanceReference
 
-    let context: SwiftNetwork.NetworkContext
-    let local: Endpoint
-    let remote: Endpoint
+    let context: NetworkContext
 
     /// Attaches a new outbound flow for a stream.
     ///
@@ -48,8 +46,8 @@ struct QUICStreamOpener {
 
         return try self.listener.invokeAttachUpperStreamProtocolToNewFlow(
             reference,
-            remote: self.remote,
-            local: self.local,
+            remote: nil,
+            local: nil,
             parameters: parameters,
             path: path
         )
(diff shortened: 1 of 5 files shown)


Sources/NIOQUIC/Streams/Transport/QUICStreamOpener.swift
```diff
@@ -0,0 +1,57 @@
...
+struct QUICStreamOpener {
+    /// The connection's stream listener.
+    let listener: StreamListenerLinkage
+
+    /// The stack's connection instance.
+    let connection: ProtocolInstanceReference
+
+    let context: SwiftNetwork.NetworkContext
+    let local: Endpoint
+    let remote: Endpoint
```
josephnoir: Given that a stream is related to a connection, do we need these or can we just pass nil and avoid these?

Sources/NIOQUIC/Streams/Transport/QUICStreamTable.swift
```diff
@@ -45,6 +45,10 @@ final class QUICStreamTable<Consumer: QUICStreamConsumer & ~Copyable> {
     @usableFromInline
     var _outputPending: Bool
 
+    /// Used for attaching locally opened streams to the `SwiftNetwork` stack. Returns `nil` if
+    /// the table isn't associated with a connection.
```
josephnoir:
```suggestion
    /// Used for attaching locally opened streams to the `SwiftNetwork` stack. `nil` if
    /// the table isn't associated with a connection.
```
🤔

Sources/NIOQUIC/Streams/Transport/QUICStreamOpener.swift
glbrntt: Good question. We currently pass them in https://github.com/apple/swift-nio-quic/blob/9e2664730fe71d154e13c27d35d74703ed16558c/Sources/NIOQUIC/SwiftNetwork/QUICChannelStreamHandler.swift#L207-L213

I chased them up the stack and it looks like they aren't required. I removed them in the existing path and tests still pass so I'll remove them. Good spot!

josephnoir (approved):
Nice one!

PR #100, opened 2026-09-10
Add a bunch of generics

glbrntt:
Motivation:

The stream table relies on a generic consumer parameter. We need space in the stack to propagate that through.

Modifications:

Wire up the `QUICStreamConsumer` as a generic parameter throughout the stack.

This is mostly a mechanical change with a few notable exceptions:
- `QUICHandler` becomes `QUICHandler<Consumer>`
- Add a `QUICStreamChannels` consumer which is currently an uninhabited type, this is the consumer for when we use the `Channel` path for streams
- Move some of the `QUICHandler` factory methods to an extension where the consumer is the `QUICStreamChannels` type (which makes them slightly easier to use).
- A few nested types which didn't rely on generics got moved out of the containing type (to avoid unnecessary generics)

Result:

Generics are in place so that we can plug in the stream table.


Sources/NIOQUIC/ConnectionRegistry.swift
```diff
@@ -131,8 +131,23 @@ struct ConnectionRegistry<Value> {
...
-extension ConnectionRegistry where Value == QUICConnectionChannel.TransportView {
+protocol QUICConnectionRegistryView {
+    func shutdown(promise: EventLoopPromise<Void>?)
+    func forceClose()
+    func parentChannelInactive()
+    func parentChannelWritabilityChanged(to writable: Bool)
+    func parentChannelUserInboundEventTriggered(_ event: Any)
+}
+
+/// A class because a local `var` trips up CMO.
```
josephnoir, line 144: 👀

josephnoir (approved):

glbrntt:
API breakage is expected, merging over it.

PR #101, opened 2026-09-11
Don't crash on inbound flows delivered after teardown

glbrntt:
Motivation:

9e266473 changed I/O slightly which surfaced a bug in the new flow handler.

Detaching stops new events from being routed to the handler but doesn't cancel ones that are already queued. `teardown()` is a direct call so it skips the event queue. A stream opened in the same read as the close is delivered after teardown where the lower protocol a `nil` event state index (so traps).

`close()` made it easy to hit: it called `teardown()` directly rather than going via `tearDownConnectionState()`, which is what clears `connectionNewFlowHandler`.

Modifications:

- Make lower protocol optional and `nil` it out on teardown, dropping inbound flows which arrive after that
- Only tear down from `tearDownConnectionState()`
- Clean up synchronously in `close()` when `stop()` wasn't called, nothing else will

Result:

Fewer crashes


josephnoir (approved):
Okay to fix the bug, but I'm wondering if we can combine `hasEstablishedConnection` and `didStop` into a single `shouldTearDownConnection`. That hinges on the question if `newFlowHandler.stop(error:)` makes the call to `tearDownConnectionState` state for `!hasEstablishedConnection` redundant or if it is needed anyway.

PR #105, opened 2026-09-14
Hold undelivered reads in a FrameArray

glbrntt:
Motivation:

Frame is large, and moving it is expensive. QUICStreamCore stored frames consumes a `FrameArray` and stores the frames in a `UniqueDeque`. Instead of moving the frame, we should try to hold on to the `FrameArray` where possible, i.e. use it instead of `UniqueDeque` and avoid the copy if possible.

Modifications:

- Hold 'undeliveredReads' as a 'FrameArray' and fill it with 'add(frames:)' rather than a frame-at-a-time append.
- Remove the 'UniqueDeque<Frame>' helpers and use the 'FrameArray' equivalents
- Add some more unit tests

Result:

Fewer copies


Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
```diff
@@ -777,6 +790,19 @@ extension QUICStreamCore {
...
+extension QUICStreamCore {
+    /// Adds `frames` to the undelivered reads, as though the stack had just handed them over.
+    mutating func _forTesting_addUndeliveredReads(_ frames: consuming FrameArray) {
+        self.undeliveredReads.add(frames: frames)
+    }
+
+    /// Whether any frame the stack handed over is still undelivered.
+    var _forTesting_hasUndeliveredReads: Bool {
+        !self.undeliveredReads.isEmpty
+    }
```
josephnoir, line 804: We have an `#if DEBUG` around testing functions in `SwiftNetworkQUICConnection.swift`, can we do the same here?

glbrntt: Do we actually want to do that? It's not a given that tests always run in debug mode. I don't have a strong preference either way but if we use `#if DEBUG` we should also wrap the tests that use this API.

josephnoir: It's nice to know the API is only used in tests and the release builds without it. At the moment we only do release builds, but not tests. Enabling release tests is a reasonable argument, but we should be consistent about it.

glbrntt: Okay. I'll add `#if DEBUG` here and to the tests as well.

josephnoir (approved):
Thanks!

PR #110, opened 2026-09-14
Refuse a flush before the stream has an ID

glbrntt:
Motivation:

A locally opened stream can be blocked on the peer's MAX_STREAMS credit. Until the stream becomes 'connected' it won't have an ID and writing data into SwiftNetwork will throw an error.

At the moment 'flush(fin:)' incorrectly swallows that error.

Modifications:

- Treat '.notConnected' as a refusal and don't drop the writes. This lets a caller try to write once the stream becomes connected (i.e. gets the '.opened') event.
- Fix the tests which were missing the `@Test` annotation...

Result:

Flushes are refused if trying too early


Sources/NIOQUIC/Streams/Transport/QUICStreamCore.swift
```diff
@@ -577,13 +578,16 @@ extension QUICStreamCore {
...
     /// - Throws: If the send side can no longer carry what was queued, in which case the queue is
-    ///   dropped; if the stack rejected the write; or if `fin` was asked for before the stack
+    ///   dropped; if the stack rejected the write; or if the flush happened before the stack
     ///   assigned the stream its ID, in which case the queue is left as it was so the flush can be
     ///   repeated.
     @usableFromInline
     mutating func flush(fin: Bool) throws {
         if self.pendingWrites.isEmpty && !fin { return }
 
+        // Refusals drop outbound writes unless there stream doesn't yet have an ID. Callers may
```
josephnoir:
```suggestion
        // Refusals drop outbound writes unless the stream doesn't yet have an ID. Callers may
```

josephnoir (approved):

PR #112, opened 2026-09-15
Add tests and make createOutboundConnection public

glbrntt:
Motivation:

The table backed streams have few integration tests: we should have more

Modifications:

- Add a bunch of integration tests, equivalent to existing integration tests where possible
- Make 'createOutboundConnection' public: it's used by tests and should be public anyway
- Fix a few docs

Result:

Better test coverage


Tests/IntegrationTests/StreamTableTests.swift
```diff
@@ -0,0 +1,903 @@
...
+            }
+            #expect(visits.last?.bytesReadSoFar == ByteBuffer(string: "<b>Success</b>"))
+        }
+
+        #expect(failures.withLockedValue { $0 }.isEmpty)
+    }
+
+    @available(anyAppleOS 26, *)
+    @Test
+    func testReceivedStopSendingWithOptInHalfClosesOutput() async throws {
```
josephnoir: IIRC, this API always supports half-closing. Did we document that difference somewhere?

```suggestion
    func testReceivedStopSendingHalfClosesOutput() async throws {
```

josephnoir (approved):
Looks great, and I like the test harness!

Tests/IntegrationTests/StreamTableTests.swift
glbrntt: Ah, good catch, that'll teach me to blindly copy test names 😅. 

With the documentation: should it be documented on the lightweight streams? The lightweight streams are the "full" API in some sense and it's the `Channel` based API that needs opting in to half-closure so it'd make more sense for it to be documented instead. WDYT?

josephnoir: Good point. Maybe should update the docs on the channel option to clarify?

glbrntt: I think it's quite well documented already?

https://github.com/apple/swift-nio-quic-helpers/blob/9f02d3052753ddfd6ad0ced39c2fc2ec275817ce/Sources/NIOQUICHelpers/ChannelOption.swift#L32-L44

josephnoir: I'm thinking of someone seeing that channel option and wondering where the equivalent for the QUICStreams API is. And given that it is not available, they might also wonder what the default is.

PR #114, opened 2026-09-18
Support for connection limits

josephnoir:
### Motivation

Services might want to limit the connections they handle concurrently to keep within expectation. This PR introduces three different configuration options:

1. Total connection limit: Limits all connections in any state.
2. Handshake limit: Limits connections currently in the handshake state.
3. Handshake rate limiting: Limits the frequency of accepting new connections.

Total connections is an overall limit that will bound both other limits as well. Similarly once the handshake limit is reached no new connections can go into handshake state independent of the current rate limit.

### Modifications

* Add an admission control to track connection and handshake counts and impose the limits.
* Add a token bucket for rate limiting support.

### Results

Services can now impose limits on connections.


Sources/NIOQUIC/ConnectionAdmissionController.swift
```diff
@@ -0,0 +1,112 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+
+/// Tacks the number of open connections and imposes limits.
```
aryan-25:
```suggestion
/// Tracks the number of open connections and imposes limits.
```

PR #115, opened 2026-09-21
Adopt a new swift-network-evolution release

josephnoir:
### Motivation

We want to benefit from the improvements that went into swift-network-evolution.

### Modifications

* Update swift-network-evolution to 0.3.0
* Add new path-related events to event dispatching.
* Adopt new signature in scheduler implementation.

### Results

Newer dependency version.


Sources/NIOQUIC/SwiftNetwork/EventLoopBackedScheduler.swift
```diff
@@ -82,10 +82,10 @@ final class EventLoopBackedScheduler: NetworkContext.Scheduler {
     /// was last scheduled with.
     func schedule(
         _ task: @escaping (() -> Void),
```
fabianfett, line 84: not a change made here, but shouldn't be the task be the last argument here?

fabianfett (approved):

Sources/NIOQUIC/SwiftNetwork/EventLoopBackedScheduler.swift
fabianfett: oh this api comes from swift-network-evolution?

